PR #56, opened 2021-07-21
typo fix

maelle:


maelle:
oops, that typo is also in #55 :-)

PR #60, opened 2021-08-31
Add missing quotes; use moar glue()

jennybc:
I am still under the influence of my recent message-overhauling activities.

### DIFF
--- a/R/resp-body.R
+++ b/R/resp-body.R
@@ -116,7 +116,7 @@ check_content_type <- function(
 
   abort(c(
     glue("Unexpected content type '{content_type}'"),
-    glue("Expecting '{type}'"),
+    glue("Expecting {type}"),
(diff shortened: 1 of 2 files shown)


R/resp-body.R
```diff
@@ -115,9 +115,9 @@ check_content_type <- function(
   }
 
   abort(c(
-    glue("Unexpected content type '{content_type}"),
-    glue("Expecting {type}"),
-    if (!is.null(suffix)) paste0("Or suffix '", suffix, "'"),
+    glue("Unexpected content type '{content_type}'"),
+    glue("Expecting '{type}'"),
```
hadley, line 119: The quotes are added above

jennybc: Yeah, I see that now that I ran / updated the snapshot tests.

hadley:
Thanks!

PR #61, opened 2021-08-31
Implement mocking

hadley:
@sckott, @nealrichardson how does this look to you?


R/req-perform.R
```diff
@@ -34,8 +38,22 @@
...
+      path = NULL,
+      verbosity = getOption("httr2_verbosity", 0L),
+      mock = getOption("httr2_mock", NULL)
+  ) {
   check_request(req)
+
+  if (!is.null(mock)) {
+    mock <- as_function(mock)
+    mock_resp <- mock(req)
+    if (!is.null(mock_resp)) {
```
nealrichardson, line 52: Why check for `NULL` specifically? If anything I'd think `stopifnot(is_response(mock_resp))` would make sense, or just let the `mock` function return whatever it wants.

R/req-perform.R
```diff
@@ -34,8 +38,22 @@
...
+      return(mock_resp)
+    }
+  }
+
```
nealrichardson, line 56: I wonder if the mock should instead be run after the next few lines since they appear to modify `req` (though I haven't looked at what they do). I vaguely remember some having to handle some subtlety in `httptest` because of default values getting set somewhere deep down that weren't a part of the `request` object earlier. It seems that here, for example, you wouldn't be able to test anything about auth or oauth tokens being set in the request because that doesn't get added to the request object until after the mock is run.

nealrichardson (commented):
Generally makes sense to me and I ❤️ how small the diff is

R/req-perform.R
hadley: I don't think `req_verbosity()` is meaningful for mocking; it's only used for debugging live requests.

And the challenge with OAuth is that signing process might need to perform its own requests (e.g. in order to refresh a token) — you'd then have to be careful to also mock those requests, which is going to be hard because they only occur occasionally.

R/req-perform.R
hadley: So the function can opt out of handling a request

nealrichardson: Ah I see you said that in the docs. Is there a use case you had in mind for this? It's not something I had considered; I'd think either you're mocking everything or nothing.

hadley: I dunno, it just seems weird to completely take over all http requests with no way to opt out.

nealrichardson: Mocking *is* weird ;) I think it would be problematic if you're mocking responses but some requests silently pass through and call out to the internet, so I can't imagine a use case for this. But I guess here you'd have to explicitly design your mock function to return NULL in those cases, so you'd have control over that?

hadley: Right. I think this design lets you opt out if you want to, but you'd have to deliberately design your function to do so.

sckott:
It looks good to me. 

I played around with it locally and it looks like it will work with webmockr and vcr (though I didn't test an actual integration with httr2).

hadley:
Thanks for all the feedback!

hadley:
BTW I'm aiming to submit the first release of httr2 to CRAN next Friday. httr2 certainly isn't done, but I think it's at a point where it makes sense for other people to try it out.

PR #69, opened 2021-09-09
Handling erroring error callback

### LINKED ISSUE #63: Think about what happens if `info` in `req_error()` fails

hadley:
Fixes #63

@lionel- does this seem right to you? I want to make sure that if someone developing an API wrapper accidentally messes up the  code for parsing addition error text it doesn't lead to a super confusing error for the user.

### DIFF
--- a/R/req-error.R
+++ b/R/req-error.R
@@ -33,5 +33,14 @@ error_is_error <- function(req, resp) {
 }
 
 error_body <- function(req, resp) {
-  req_policy_call(req, "error_body", list(resp), default = NULL)
+  tryCatch(
+    req_policy_call(req, "error_body", list(resp), default = NULL),
+    error = function(cnd) {
+      cnd <- cnd_entrace(cnd)
+      c(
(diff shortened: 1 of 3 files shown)


R/req-error.R
```diff
@@ -33,5 +33,14 @@ error_is_error <- function(req, resp) {
 }
 
 error_body <- function(req, resp) {
-  req_policy_call(req, "error_body", list(resp), default = NULL)
+  tryCatch(
+    req_policy_call(req, "error_body", list(resp), default = NULL),
+    error = function(cnd) {
+      cnd <- cnd_entrace(cnd)
```
lionel-, line 39: `cnd_entrace()` will not help much in a `tryCatch()` handler, but also you seem to return a character vector so the entracing is ignored?

lionel- (commented):
This works but I think this is a typical case where we'd use a chained error with `withCallingHandler()`, and `abort(parent = , call = )`?

We'd have to wait until rlang 1.0 until the low-level error shows up to users though.

PR #70, opened 2021-09-15
Review of "Wrapping APIs" vignette

jennybc:
In the form of a PR, as requested. I'm about to move through and comment on lots of markers I've left.

### DIFF
--- a/vignettes/wrapping-apis.Rmd
+++ b/vignettes/wrapping-apis.Rmd
@@ -70,7 +70,7 @@ resp %>% resp_body_json() %>% str()
 ### Errors
 
 It's always worth a little early experimentation to see if we get any useful information from errors.
-JENNY COMMENT HERE The httr2 details get in your way here, because if you retrieve an unsuccessful HTTP response, you automatically get an error that prevents you from further inspecting the body:
+The httr2 defaults get in your way here, because if you retrieve an unsuccessful HTTP response, you automatically get an error that prevents you from further inspecting the body:
 
 ```{r, error = TRUE, eval = faker_ok}
 req %>% 
@@ -78,7 +78,7 @@ req %>%
   req_perform()
 ```
 
-You can access the failed response with `last_response()`:
+However, you can access the last response (successful or not) with `last_response()`:
 
 ```{r, eval = faker_ok}
 resp <- last_response()
@@ -117,7 +117,7 @@ For faker, I spent a little with the [documentation](https://fakerapi.it/en#basi
 -   Every URL is of the form [`https://fakerapi.it/api/v1`](https://fakerapi.it/api/v1)`/{resource}`, and data is passed to the resource with query parameters.
     All parameters start with `_`.
 
--   Every resource has three common query parameters: `_locale`, `_quantity`, and `_seed` which have default values of `en_US`, `1`, and `null` respectively.
+-   Every resource has three common query parameters: `_locale`, `_quantity`, and `_seed` which have default values of `en_US`, `10`, and `null` respectively.
 
 -   All endpoints return JSON data.
 
@@ -161,7 +161,7 @@ So typically the next step in the process is to wrap up some individual endpoint
 
 JENNY COMMENT HERE
 
-For example, let's take the persons endpoint which has three additional parameters: `gender` (male or female), `birthday_start`, and `birthday_end`.
+For example, let's take the `persons` endpoint which has three additional parameters: `gender` (male or female), `birthday_start`, and `birthday_end`.
 A simple wrapper would start something like this:
 
 ```{r, eval = faker_ok}
@@ -233,8 +233,7 @@ has_auth <- secret_has_key("HTTR2_KEY") &&
 knitr::opts_chunk$set(eval = has_auth)
 ```
 
-Next we'll take a look at the NYTimes [Books API](https://developer.nytimes.com/docs/books-product/1/overview).
-It requires a very simple authentication with an API key that's included in every request.
+Next we'll take a look at the NYTimes [Books API](https://developer.nytimes.com/docs/books-product/1/overview). It requires a very simple authentication with an API key that's included in every request.
 When you're wrapping an API that has a key you're going to face two struggles:
 
 -   How do you test your package without sharing your key with the whole world?
@@ -293,9 +292,8 @@ So now you can understand how the following code works:
 my_key <- secret_decrypt("4Nx84VPa83dMt3X6bv0fNBlLbv3U4D1kHM76YisKEfpCarBm1UHJHARwJHCFXQSV", "HTTR2_KEY")
 ```
 
-JENNY COMMENT HERE
-
-Since most people won't have this key, n the source of this vignette I also check whether or not it's available with `has_auth <- secret_has_key("HTTR2_KEY")`, then use `knitr::opts_chunk(eval = has_auth)` so that the code is only run when possible.
+To make this vignette work everywhere (e.g. on CRAN which won't have have this key), I first figure out whether it's available with `has_auth <- secret_has_key("HTTR2_KEY")`.
+Then I do `knitr::opts_chunk(eval = has_auth)` so that all future chunks are only evaluated if the key is available.
 
 When `secret_decrypt()` is run in tests, it will automatically skip the test if the key isn't found.
 That means that your auth-requiring tests will automatically be skipped on CRAN.
@@ -346,16 +344,10 @@ resp <- request("https://api.nytimes.com/svc/books/v3") %>%
   req_perform()
 ```
 
-To see if there's any extra useful information we can again suppress the default error and inspect the result:
+To see if there's any extra useful information we can again look at `last_response():`
 
-JENNY COMMENT HERE
-
-```{r, error = TRUE}
-resp <- request("https://api.nytimes.com/svc/books/v3") %>% 
-  req_url_path_append("/reviews.json") %>% 
-  req_url_query(`api-key` = "invalid", isbn = 9780307476463) %>% 
-  req_error(is_error = ~ FALSE) %>% 
-  req_perform()
+```{r}
+resp <- last_response()
 resp
 resp %>% resp_body_json()
 ```
@@ -371,12 +363,14 @@ This should be a function that takes a response and returns a character vector o
 Once we do that and re-fetch the request, we see the additional information displayed in the R error:
 
 ```{r, error = TRUE}
+nytimes_error_body <- function(resp) {
+  resp %>% resp_body_json() %>% .$fault %>% .$faultstring
+}
+
 resp <- request("https://api.nytimes.com/svc/books/v3") %>% 
   req_url_path_append("/reviews.json") %>% 
   req_url_query(`api-key` = "invalid", isbn = 9780307476463) %>% 
-  req_error(body = function(resp) {
-    resp %>% resp_body_json() %>% .$fault %>% .$faultstring
-  }) %>% 
+  req_error(body = nytimes_error_body) %>% 
   req_perform()
 ```
 
@@ -388,8 +382,6 @@ The [frequently asked questions](https://developer.nytimes.com/faq#a11) page des
 > Yes, there are two rate limits per API: 4,000 requests per day and 10 requests per minute.
 > You should sleep 6 seconds between calls to avoid hitting the per minute rate limit.
 > If you need a higher rate limit, please contact us at [code\@nytimes.com](mailto:code@nytimes.com){.email}.
-
-JENNY SARCASTIC AND BITTER COMMENT HERE
 
 Many APIs return additional information about how long to wait when the rate limit is exceeded (often using the `Retry-After` header).
 So I deliberately violated the rate limit by quickly making 11 requests; unfortunately while the response was a standard 429 (Too many requests), it did not include any information about how long to wait in either the response body or the headers.
@@ -416,7 +408,6 @@ req <- request("https://api.nytimes.com/svc/books/v3") %>%
 ### Wrapping it up
 
 Putting together all the pieces above yields a function something like this:
-
 
 ```{r}
 nytimes_books <- function(api_key, path, ...) {
@@ -466,8 +457,6 @@ get_api_key <- function() {
 Then you could modify `nytimes_books()` to use `get_api_key()` as the default value for `api_key`.
 Since the argument is now optional, we can move it to end of the argument list, since it'll only be needed in exceptional circumstances.
 
-JENNY COMMENT HERE
-
 ```{r, eval = FALSE}
 nytimes_books <- function(path, ..., api_key = get_api_key()) {
   ...
@@ -488,8 +477,7 @@ testing_key <- function() {
 get_api_key <- function() {
   key <- Sys.getenv("NYTIMES_KEY")
   if (identical(key, "")) {
-    if (is_testing()) {
-      # JENNY COMMENT HERE
+    if (is_testing() && secret_has_key("HTTR2_KEY")) {
       testing_key()
     } else {
       stop("No API key found, please supply with `api_key` argument or with NYTIMES_KEY env var")  
@@ -568,10 +556,8 @@ resp %>% resp_body_json()
 That body looks pretty useful!
 So let's write a function that extracts the data and formats it for presentation to the user:
 
-JENNY COMMENT HERE
-
 ```{r}
-github_error_body <- function(resp) {
+gist_error_body <- function(resp) {
   body <- resp_body_json(resp)
   
   message <- body$message
@@ -580,7 +566,7 @@ github_error_body <- function(resp) {
   }
   message
 }
-github_error_body(resp)
+gist_error_body(resp)
 ```
 
 Now I can pass this function to the `body` argument of `req_error()` and it will be automatically included in the error when a request fails:
@@ -589,7 +575,7 @@ Now I can pass this function to the `body` argument of `req_error()` and it will
 request("https://api.github.com/gists") %>% 
   req_url_query(since = "yesterday") %>% 
   req_headers(Authorization = paste("token", token)) %>% 
-  req_error(body = github_error_body) %>% 
+  req_error(body = gist_error_body) %>% 
   req_perform()
 ```
 
@@ -611,11 +597,11 @@ The first tells us whether or not a response has a transient error, i.e. it's wo
 For GitHub, when the rate limit is [exceeded](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting), the response has a 403 status and a `X-RateLimit-Remaining: 0` header:
 
 ```{r}
-github_is_transient <- function(resp) {
+gist_is_transient <- function(resp) {
   resp_status(resp) == 403 && 
     resp_header(resp, "X-RateLimit-Remaining") == "0"
 }
-github_is_transient(resp)
+gist_is_transient(resp)
 ```
 
 Then we need a function tells how long to wait.
@@ -623,11 +609,11 @@ GitHub tells us when the rate limit resets (as number of seconds since 1970-01-0
 To convert that to a number of seconds to wait we first convert it to a number (since HTTP headers are always strings), then subtract off the current time (in number of seconds since 1970-01-01):
 
 ```{r}
-github_after <- function(resp) {
+gist_after <- function(resp) {
   time <- as.numeric(resp_header(resp, "X-RateLimit-Reset"))
   time - unclass(Sys.time())
 }
-github_after(resp)
+gist_after(resp)
 ```
 
 We then pass functions to `req_retry()` so httr2 has all the information it needs to handle rate-limiting automatically:
@@ -648,7 +634,7 @@ You also need to supply either `max_tries` or `max_seconds` in order to activate
 Let's wrap up everything we've learned so far into a single function that creates a request:
 
 ```{r}
-req_github <- function(token) {
+req_gist <- function(token) {
   request("https://api.github.com/gists") %>% 
     req_headers(Authorization = paste("token", token)) %>% 
     req_error(body = github_error_body) %>% 
@@ -659,7 +645,7 @@ req_github <- function(token) {
 }
 
 # Check it works:
-req_github(token) %>% 
+req_gist(token) %>% 
   req_perform()
 ```
 
@@ -667,13 +653,11 @@ We'll use this as the basis to solve the next challenge: uploading a gist.
 
 ### Sending data
 
-To [create a gist](https://docs.github.com/en/rest/reference/gists#create-a-gist) we need to send a `POST` request with a body that contains data encoded as JSON.
+To [create a gist](https://docs.github.com/en/rest/reference/gists#create-a-gist) we need to change the method to `POST` and add a body that contains data encoded as JSON.
 httr2 provides one function that does both of these things: `req_body_json()`:
 
-JENNY COMMENT HERE
-
 ```{r}
-req <- req_github(token) %>% 
+req <- req_gist(token) %>% 
   req_body_json(list(
     description = "This is my cool gist!",
     files = list(test.R = list(content = "print('Hi!')")),
@@ -702,7 +686,7 @@ To do so, I need to again send JSON encoded data, but this time I need to use th
 So after adding the data to request, I use `req_method()` to override the default method:
 
 ```{r}
-req <- req_github(token) %>% 
+req <- req_gist(token) %>% 
   req_url_path_append(id) %>% 
   req_body_json(list(description = "This is a lame gist")) %>% 
   req_method("PATCH")
@@ -714,7 +698,7 @@ req %>% req_dry_run()
 Deleting a gist is similar, except we don't send any data, we just need to adjust the default method from `GET` to `DELETE`.
 
 ```{r}
-req <- req_github(token) %>% 
+req <- req_gist(token) %>% 
   req_url_path_append(id) %>% 
   req_method("DELETE")
 req %>% req_dry_run()
@@ -786,7 +770,8 @@ This flow can't be used inside a vignette because it's designed specifically for
 
 Other flows provide different ways of getting the token:
 
--   `req_oauth_client_credentials()` is used to allow the client to perform actions on its own behalf (instead of on behalf of some other user). JENNY COMMENT HERE
+-   `req_oauth_client_credentials()` is used to allow the client to perform actions on its own behalf (instead of on behalf of some other user).
+    JENNY COMMENT HERE
 
 -   `req_oauth_device()` uses the "device" flow which is designed for devices like TVs that don't have an easy way to enter data.
     It also works well from the console.
@@ -833,16 +818,7 @@ request("https://api.github.com/user") %>%
 ```
 
 This is important because most APIs provide only a short-lived access token that needs to be regularly refreshed using a longer-lived refresh token.
-httr2 will automatically refresh the token in two cases:
-
--   The token includes an expiry date that's in the past.
-
--   The request returns a 401 status code.
-
-JENNY COMMENT HERE
-
-If the access token has expired, httr2 will automatically retry automatically refresh before the first request.
-If the request errors with a 401 and `invalid_token` error in the `WWW-authenticate` header.
+httr2 will automatically refresh the token if its expired (i.e. its expiry date is in the past) or if the request errors with a 401 and there's an `invalid_token` error in the `WWW-authenticate` header.
 
 ### Caching
 
--- a/vignettes/wrapping-apis.Rmd
+++ b/vignettes/wrapping-apis.Rmd
@@ -97,16 +97,14 @@ We'll have another go later with an API that does provide more details.
 
 ### User agent
 
-When making a package that calls an API it's considered polite to set a user agent, so that, if your app accidentally does something horribly wrong, the developers of the website can figure out who to reach out to.
+If you're wrapping this code into a package, it's considered polite to set a user agent, so that, if your app accidentally does something horribly wrong, the developers of the website can figure out who to reach out to.
 You can do this with the `req_user_agent()` function:
 
 ```{r, eval = faker_ok}
 req %>%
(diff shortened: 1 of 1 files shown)


vignettes/wrapping-apis.Rmd
```diff
@@ -397,11 +417,13 @@ req <- request("https://api.nytimes.com/svc/books/v3") %>%
 
 Putting together all the pieces above yields a function something like this:
 
+
 ```{r}
 nytimes_books <- function(api_key, path, ...) {
   request("https://api.nytimes.com/svc/books/v3") %>% 
     req_url_path_append("/reviews.json") %>% 
     req_url_query(..., `api-key` = api_key) %>% 
+    # JENNY COMMENT HERE
```
jennybc, line 426: I see you do model this in a later example, but I wonder if you want to predefine the error-body-handling function? Either outside of `nytimes_books()` or inside. Because I can't imagine these would ever stay simple enough to define inline like this. Also this feels pretty hostile to writing tests.

vignettes/wrapping-apis.Rmd
```diff
@@ -416,19 +438,21 @@ drunk$results[[1]]$summary
...
 You'd also want to provide some convenient way for the user to supply their own API key.
 
-### User supplied key
+### User-supplied key
 
 A good place to start is an environment variable, because environment variables are easy to set without typing anything in the console (which can get accidentally shared via your `.Rhistory`) and are easily set in automated processes.
 Then you'd write a function to retrieve the API key, returning a helpful message if it's not found:
 
+JENNY COMMENT HERE
+
```
jennybc, line 455: Can't remember what I wanted to say here.

vignettes/wrapping-apis.Rmd
```diff
@@ -442,6 +466,8 @@ get_api_key <- function() {
 Then you could modify `nytimes_books()` to use `get_api_key()` as the default value for `api_key`.
 Since the argument is now optional, we can move it to end of the argument list, since it'll only be needed in exceptional circumstances.
 
+JENNY COMMENT HERE
```
jennybc, line 469: Maybe also worth pointing out "to the end and *after `...`*". If indeed that is intentional, but I suspect it is.

vignettes/wrapping-apis.Rmd
```diff
@@ -463,6 +489,7 @@ get_api_key <- function() {
   key <- Sys.getenv("NYTIMES_KEY")
   if (identical(key, "")) {
     if (is_testing()) {
+      # JENNY COMMENT HERE
```
jennybc, line 492: From the code I'm seeing here, it's not obvious to me that this condition (`is_testing()`) is sufficient to guarantee that `testing_key()` will successfully return a key. Have I missed something?

vignettes/wrapping-apis.Rmd
```diff
@@ -485,13 +512,16 @@ A token is an alternative to a username and password.
 You have one username + password per site; you can have one token per use case.
 This lets each use case have a minimal set of permissions, and you can easily revoke one token without affecting any other use case.
 
-I created a personal access token specifically for this vignette that can only access gists, and as in the last example, stored an encrypted version in this vignette:
+JENNY COMMENT HERE
```
jennybc, line 515: Can't remember what I wanted to say here.

vignettes/wrapping-apis.Rmd
```diff
@@ -538,6 +568,8 @@ resp %>% resp_body_json()
 That body looks pretty useful!
 So let's write a function that extracts the data and formats it for presentation to the user:
 
+JENNY COMMENT HERE
```
jennybc, line 571: It feels like all these functions should probably be prefixed with `gist_` instead of `github_`.

Also, kind of like rate limiting, I have yet to work with an API that structures its errors consistently enough to really do this. That would of course be great! I wonder how typical this 🤔 Of course, I know I will be able to take my extremely ugly version of this for Google errors and insert them into this framework. So what I'm really wondering is what these error-body-processing functions will really look like in the wild.

vignettes/wrapping-apis.Rmd
```diff
@@ -635,9 +667,11 @@ We'll use this as the basis to solve the next challenge: uploading a gist.
 
 ### Sending data
 
-To [create a gist](https://docs.github.com/en/rest/reference/gists#create-a-gist) we need to send a `POST` request with a body that contains data encodedd as JSON.
+To [create a gist](https://docs.github.com/en/rest/reference/gists#create-a-gist) we need to send a `POST` request with a body that contains data encoded as JSON.
 httr2 provides one function that does both of these things: `req_body_json()`:
 
+JENNY COMMENT HERE
```
jennybc, line 673: At first, I understood "does both of those things" to suggest that `req_body_json()` does both of these things:
* **send** a `POST` request
* with a body ...

and I was surprised that it **sent** a request. But I was just confused. It only **forms** a `POST` request. Maybe this could be reworded.

vignettes/wrapping-apis.Rmd
```diff
@@ -689,6 +723,8 @@ req %>% req_perform()
 
 ## OAuth
 
+JENNY PEDANTIC COMMENT HERE
```
jennybc, line 726: I know the whole authentication vs. authorization thing and that we're sort of talking about both. But if we're going to pick one of those words to feature in this paragraph, here authorization somehow feels like the better choice?

vignettes/wrapping-apis.Rmd
```diff
@@ -698,13 +734,15 @@ You've almost certainly used this before on the web; it's used in most cases whe
 
 ### Clients
 
-The first step in working with any OAuth api is to create a client.
-This involves you registering for a developer account on the APIs website and creating a new OAuth app.
+The first step in working with any OAuth API is to create a client.
+This involves you registering for a developer account on the API's website and creating a new OAuth app.
 The process varies from API to API, but at the end of it you'll get a client id and in most cases a client secret.
 
+JENNY COMMENT HERE
```
jennybc, line 741: I feel like the scenario of "user provides own client" is not on the radar here at all, but it should be.

vignettes/wrapping-apis.Rmd
```diff
@@ -744,11 +782,11 @@ You can try it out by running this code:
...
 ```
 
-This flow can't be used inside a vignette because it's designed specifically for interactive use: it will open a webpage on Github that requires you to interactively confirm it's ok for this app to use your GitHub account.
+This flow can't be used inside a vignette because it's designed specifically for interactive use: it will open a webpage on GitHub that requires you to interactively confirm it's OK for this app to use your GitHub account.
 
-Other flows provide different ways getting the token:
+Other flows provide different ways of getting the token:
 
--   `req_oauth_client_credentials()` is used to allow the client to perform actions on its own behalf (instead of on behalf of some other user).
+-   `req_oauth_client_credentials()` is used to allow the client to perform actions on its own behalf (instead of on behalf of some other user). JENNY COMMENT HERE
```
jennybc, line 789: I perceive the term "service account (token)" to be common enough to mention here. That's what I think this is about, yeah?

vignettes/wrapping-apis.Rmd
```diff
@@ -799,20 +837,22 @@ httr2 will automatically refresh the token in two cases:
 
 -   The token includes an expiry date that's in the past.
 
--   The request returns a 401 status code and
+-   The request returns a 401 status code.
+
+JENNY COMMENT HERE
```
jennybc, line 842: There seems to be some garbled / truncated wording in the bullet list above and the following 2 sentences.

vignettes/wrapping-apis.Rmd
```diff
@@ -799,20 +837,22 @@ httr2 will automatically refresh the token in two cases:
...
 In some cases, you may want to save the token so that it's automatically used across sessions.
-This is easy to do (just set `cache_disk = TRUE` in `req_oauth_auth_code()`) but you need to carefully consider the consequences of saving the users credentials on disk.
+This is easy to do (just set `cache_disk = TRUE` in `req_oauth_auth_code()`) but you need to carefully consider the consequences of saving the user's credentials on disk.
 
 httr2 does the best it can to save these credentials securely.
 They are stored in a local cache directory (`rappdirs::user_cache_dir("httr2"))` that should only be accessible to the current user, and are encrypted so they will be hard for any package other than httr2 to read.
-However, there's no way to protect another R code from using httr2 to access them, so if you do choose to cache tokens, you should inform the user and give them the ability to opt-out.
+However, there's no way to protect other R code from using httr2 to access them, so if you do choose to cache tokens, you should inform the user and give them the ability to opt-out.
```
jennybc, line 855:
```suggestion
However, there's no way to prevent other R code from using httr2 to access them, so if you do choose to cache tokens, you should inform the user and give them the ability to opt-out.
```

I saw this mis-wording once I was reviewing on GitHub.

vignettes/wrapping-apis.Rmd
```diff
@@ -262,17 +272,19 @@ This is a two step process:
...
 
 Now you can encrypt your API key and store it in your tests --- there's no way for anyone to decrypt it unless they have your key.
 
+JENNY COMMENT HERE
+
 ```{r}
-x <- secret_encrypt("my-secret", "HTTR2_KEY")
-x
-secret_decrypt(x, "HTTR2_KEY")
+x_enc <- secret_encrypt("API-key-from-NYTimes-Books", "HTTR2_KEY")
```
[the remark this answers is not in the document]
hadley: Yup

vignettes/wrapping-apis.Rmd
```diff
@@ -281,24 +293,28 @@ So now you can understand how the following code works:
...
 
-All in all, this mean you need to be a little cautious if you need to get help with a reprex; you'll need to manually redact the key.
+All in all, this means you need to be a little cautious if you need to get help with a reprex; you'll need to manually redact the key.
 
 ### Basic request
 
-Now lets perform a test request and look at the response:
+Now let's perform a test request and look at the response:
+
+JENNY COMMENT HERE
```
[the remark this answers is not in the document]
hadley: Ooops! I'll have to think about how to avoid this? Maybe some `req_redact()` function that's automatically applied to the output of `req_dryrun()`? 

Or maybe I should just embrace it, with a comment.

vignettes/wrapping-apis.Rmd
```diff
@@ -332,6 +348,8 @@ resp <- request("https://api.nytimes.com/svc/books/v3") %>%
 
 To see if there's any extra useful information we can again suppress the default error and inspect the result:
 
+JENNY COMMENT HERE
+
```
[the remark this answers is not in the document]
hadley: I figured it out — I should be using `last_request()` as in the previous example.

vignettes/wrapping-apis.Rmd
```diff
@@ -371,10 +389,12 @@ The [frequently asked questions](https://developer.nytimes.com/faq#a11) page des
 > You should sleep 6 seconds between calls to avoid hitting the per minute rate limit.
 > If you need a higher rate limit, please contact us at [code\@nytimes.com](mailto:code@nytimes.com){.email}.
 
+JENNY SARCASTIC AND BITTER COMMENT HERE
```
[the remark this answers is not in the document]
hadley: I think it's getting more common, although even GitHub doesn't do it exactly according to the spec.

vignettes/wrapping-apis.Rmd
hadley: Good call. I do that later on, but not here.

vignettes/wrapping-apis.Rmd
hadley: I was just using `...` casually here, but yes, I would recommend putting it after `...` if it exists (i.e. https://design.tidyverse.org/dots-position.html). Do you think I need to explicitly call that out here?

vignettes/wrapping-apis.Rmd
hadley: Nope! I forgot a `secret_has_key("HTTR2_KEY")`.

vignettes/wrapping-apis.Rmd
hadley: Agreed.

Yeah, and in my experience typically there's no way to find out how each endpoint actually structures its errors without deliberately forcing them. I'll add a note to that effect.

vignettes/wrapping-apis.Rmd
hadley: Good point!

vignettes/wrapping-apis.Rmd
hadley: You mean explicitly connect this function to "service" accounts" or explain more what a service account is?

hadley (commented):
Thanks! Responded/fixed a bunch of the smaller stuff, and I'll tackle the bigger stuff later.

vignettes/wrapping-apis.Rmd
```diff
@@ -105,21 +106,25 @@ req %>%
   req_dry_run()
 ```
 
+JENNY COMMENT HERE
```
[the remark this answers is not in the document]
jennybc: Oh, no, I don't think you should change it.

I had gotten the impression there was *no* default user agent and one always had to set it. I think this just needs a sentence somewhere that, when you create an API-specific wrapper package, part of the infrastructure built into your main "request constructor" should include setting the user agent.

vignettes/wrapping-apis.Rmd
```diff
@@ -105,21 +106,25 @@ req %>%
...
-Once you've successful made a few requests, it's worth seeing if you can figure out the general pattern so you can wrap it up into a function that will become the core of your package.
+Once you've made a few successful requests, it's worth seeing if you can figure out the general pattern so you can wrap it up into a function that will become the core of your package.
 
 For faker, I spent a little with the [documentation](https://fakerapi.it/en#basic-usage) noting some commonalities:
 
 -   Every URL is of the form [`https://fakerapi.it/api/v1`](https://fakerapi.it/api/v1)`/{resource}`, and data is passed to the resource with query parameters.
     All parameters start with `_`.
 
--   Every resource has three common query parameters: `_locale`, `_quantity`, and `_seed` which have default values of `en_US`, `1` and `null` respectively.
+-   Every resource has three common query parameters: `_locale`, `_quantity`, and `_seed` which have default values of `en_US`, `10`, and `null` respectively.
```
jennybc, line 120: So the default for `_quantity` is `10` not `1`? If so, that needs to change in a few other places.

vignettes/wrapping-apis.Rmd
```diff
@@ -143,6 +148,8 @@ str(faker("images", width = 300))
 
 There's one trick to notice here --- `req_url_query()` uses [dynamic dots](https://rlang.r-lib.org/reference/dyn-dots.html), so we can use `!!!` to convert (e.g.) `` req_url_query(req, !!!list(`_quantity` = 1, `_locale` = "en_US") `` into `` req_url_query(req, `_quantity` = 1, `_locale` = "en_US") ``.
 
+JENNY COMMENT HERE
```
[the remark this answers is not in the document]
jennybc: OK I guess was reading this as if you were demo'ing with `faker()`, which has this `_` prefixing business. But you actually just use `req_url_query()`, so the claim does make sense.

vignettes/wrapping-apis.Rmd
```diff
@@ -152,6 +159,8 @@ If the API you are wrapping is more variable, you may want to return a request o
 But it's not very friendly for the package user who might not know anything about web APIs.
 So typically the next step in the process is to wrap up some individual endpoints with their own functions.
 
+JENNY COMMENT HERE
+
 For example, let's take the persons endpoint which has three additional parameters: `gender` (male or female), `birthday_start`, and `birthday_end`.
```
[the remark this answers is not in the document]
jennybc: Oh, right, we're letting The API randomly generate a birthday.

vignettes/wrapping-apis.Rmd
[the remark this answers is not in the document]
jennybc: That sounds right!

vignettes/wrapping-apis.Rmd
jennybc: Well, I really like this principle, so I'd take the chance to publicize it. But I also don't think matters much here. It's tangential.

vignettes/wrapping-apis.Rmd
jennybc: And, with most APIs, it can be very hard to intentionally cause errors and capture them for study. You basically have to wait for yourself or a user to experience it and then respond.

vignettes/wrapping-apis.Rmd
jennybc: Yeah, I just mean

> explicitly connect this function to "service" accounts"

vignettes/wrapping-apis.Rmd
hadley: The default for the API is 10; but I want to change it to 1 for these demos.

vignettes/wrapping-apis.Rmd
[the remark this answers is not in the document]
hadley: Ok, I'll just make it more clear that is in a fill in the blanks example.

vignettes/wrapping-apis.Rmd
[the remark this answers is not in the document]
hadley: I've added an explanation of why I do it — here the motivation is primarily that it's a pain to type argument names that start with `_` since you have to backtick quote them.

vignettes/wrapping-apis.Rmd
```diff
@@ -105,21 +106,25 @@ req %>%
...
     All parameters start with `_`.
 
--   Every resource has three common query parameters: `_locale`, `_quantity`, and `_seed` which have default values of `en_US`, `1` and `null` respectively.
+-   Every resource has three common query parameters: `_locale`, `_quantity`, and `_seed` which have default values of `en_US`, `1`, and `null` respectively.
 
 -   All endpoints return JSON data.
 
 This lead me to construct the following function:
 
+JENNY COMMENT HERE
```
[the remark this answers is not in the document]
hadley: Ok, I've explained this a little more now.  Also explained what `NULL` does here.

vignettes/wrapping-apis.Rmd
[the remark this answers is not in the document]
hadley: Decided to embrace it

vignettes/wrapping-apis.Rmd
hadley: Ooops, no I didn't — `secret_decrypt()` will automatically skip if `HTTR2_KEY` is not found.

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
+## Secret management
 
-```{r include = FALSE}
-has_auth <- secret_has_key("HTTR2_KEY") && 
-  # Don't run in R CMD check on GHA because it hits NYTimes rate limits
-  !identical(Sys.getenv("GITHUB_WORKFLOW"), "R-CMD-check")
-knitr::opts_chunk$set(eval = has_auth)
-```
+We need to take a quick break from APIs to talk about secrets.
+Secrets are important, because every API (except for very simnple APIs like faker) is going to require that you identify yourself in some way typically with a username and password or an API key.
```
jennybc, line 232:
```suggestion
Secrets are important, because every API (except for very simnple APIs like faker) is going to require that you identify yourself in some way, typically with an API key or a token.
```

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
+And even if you expect that your users will usually provide this information, you're still going to need your own credentials in order to actually test your package.
 
-Next we'll take a look at the NYTimes [Books API](https://developer.nytimes.com/docs/books-product/1/overview). It requires a very simple authentication with an API key that's included in every request.
-When you're wrapping an API that has a key you're going to face two struggles:
+This system is probably overkill if you only have one secret that you need to share in one or two places.
+But you almost invariably accumulate more secrets over time, and more people and computers that you need to share it with, so I think spending little time to understand this system and set up it for your package will pay off in the long term.
```
jennybc, line 236:
```suggestion
But you almost invariably accumulate more secrets over time, and more people and computers that you need to share them with, so I think spending a little time to understand this system and set up it for your package will pay off in the long term.
```

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
 
-```{r}
-secret_make_key()
-```
+2.  You scramble your secrets with `secret_encrypt()` and store the resulting text directly in the source code of your package:
 
-(Don't copy and paste this key; run `secret_make_key()` yourself to make your own personal key.)
+    ```{r}
+    secret <- secret_encrypt("this is a secret message", key)
+    secret
```
jennybc, line 256:
```suggestion
    secret_scrambled <- secret_encrypt("this is a secret message", key)
    secret_scrambled
```

I know this is a bit pedantic but I seriously think it helps when you're first absorbing all this.

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
+    ```
 
-Now you need to make this key available in the places were you want to unlock the secrets.
-httr2 provides tools to retrieve keys from environment variables because this is a common way of communicating these sort of secrets.
-So start by opening your `.Renviron` (e.g. with `usethis::edit_r_environ()`) and add a line like:
+3.  When needed, you descramble the the secret using `secret_decrypt()`:
+
+    ```{r}
+    secret_decrypt(secret, key)
```
jennybc, line 262:
```suggestion
    secret_decrypt(secret_scrambled, key)
```

jennybc: It feels like a concrete example is needed around here, but maybe that gets taken care of soon enough. Specifically, the code above makes it look like you'll store the scrambled version in a variable, but IRL these will be string literals.

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
+    ```
 
-    YOURPACKAGE_KEY=key_that_you_generated_above
+### Package secrets
 
-Another place you'll probably want to provide the key is in your Github actions.
-This is a two step process:
+You can create an number of encryption keys, but I highly recommend that you create one key per package, which I'll call the **package** key.
```
jennybc, line 267:
```suggestion
You can create any number of encryption keys, but I highly recommend that you create one key per package, which I'll call the **package** key.
```

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
+httr2 is built around the notion that this key should live in an environment variable.
+So the first step is to make your package key available on your local development machine by adding a line to your your user-level `.Renviron` (which you can easily open with `usethis::edit_r_environ()`):
+
+    YOURPACKAGE_KEY=key_you_generated_with_secret_make_key
+
+```{r, include = FALSE}
+Sys.setenv(YOURPACKAGE_KEY = secret_make_key())
+```
+
+Now (after you restart R), you'll be able to take advantage of a special `secret_encrypt()` and `secret_decrypt()` feature: the encryption key can be the name of an environment variable.
```
jennybc, line 279:
```suggestion
Now (after you restart R), you'll be able to take advantage of a special `secret_encrypt()` and `secret_decrypt()` feature: the `key` argument can be the name of an environment variable, instead of the encryption key itself.
In fact, this is most natural usage.
```

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
+
+```{r}
+secret <- secret_encrypt("my personal secret", "YOURPACKAGE_KEY")
+secret
+secret_decrypt(secret, "YOURPACKAGE_KEY")
```
jennybc, line 284:
```suggestion
secret_scrambled <- secret_encrypt("my personal secret", "YOURPACKAGE_KEY")
secret_scrambled
secret_decrypt(secret_scrambled, "YOURPACKAGE_KEY")
```

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
+```
+
+You'll also need to make the key available in your GitHub Actions so your automated tests can use it.
```
jennybc, line 287: At any of these points, do we want to also point out the value of this whole system re: documentation? I.e. it make it clear you probably want to expose the encryption key to your pkgdown workflow, in addition to your check workflows.

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
+This requires two steps:
 
 1.  Add the key to your [repository secrets](https://docs.github.com/en/actions/reference/encrypted-secrets).
 
 2.  Share the secret with the workflows that need it by adding a line to the appropriate workflow:
```
jennybc, line 292:
```suggestion
2.  Share the key with the workflows that need it by adding a line to the appropriate workflow:
```

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
         env:
-          HTTR2_KEY: ${{ secrets.HTTR2_KEY }}
+          YOURPACKAGE_KEY: ${{ secrets.YOURPACKAGE_KEY }}
     ```
 
-    You can see how httr2 does it in <https://github.com/r-lib/httr2/blob/master/.github/workflows/R-CMD-check.yaml>.
+    You can see how httr2 does it in [its GitHub workflow](https://github.com/r-lib/httr2/blob/master/.github/workflows/R-CMD-check.yaml){.uri}.
 
-Now you can encrypt your API key and store it in your tests --- there's no way for anyone to decrypt it unless they have your key.
+Other continuous integration platforms will offer similar ways to make secrets available in environment variables.
```
jennybc, line 301:
```suggestion
Other continuous integration platforms will offer similar ways to make secrets available in secure environment variables.
```

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
 
-```{r}
-x <- secret_encrypt("my-secret", "HTTR2_KEY")
-x
-secret_decrypt(x, "HTTR2_KEY")
+### When the package key isn't available
+
+There's an important computer that won't have your package key: CRAN.
```
jennybc, line 305: Is it also worth mentioning that this applies to PRs from external forks? If you don't do this, you will never get any decent info from CI for such PRs. Whereas with this, assuming you have some tests that don't require live API access, at least those tests will provide info and the others will be skipped vs. failing.

vignettes/wrapping-apis.Rmd
jennybc: I'm not totally happy with "natural" but I want to convey "most intended", "most planned for", "should be most common".

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,130 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
-httr2 provides tools to retrieve keys from environment variables because this is a common way of communicating these sort of secrets.
-So start by opening your `.Renviron` (e.g. with `usethis::edit_r_environ()`) and add a line like:
+3.  When needed, you descramble the the secret using `secret_decrypt()`:
+
+    ```{r}
+    secret_decrypt(secret, key)
+    ```
 
-    YOURPACKAGE_KEY=key_that_you_generated_above
+### Package secrets
```
jennybc, line 265:
```suggestion
### Package keys and secrets
```

vignettes/wrapping-apis.Rmd
```diff
@@ -213,92 +224,131 @@ faker_person <- function(gender = NULL, birthday_start = NULL, birthday_end = NU
...
+So the first step is to make your package key available on your local development machine by adding a line to your your user-level `.Renviron` (which you can easily open with `usethis::edit_r_environ()`):
+
+    YOURPACKAGE_KEY=key_you_generated_with_secret_make_key
+
+```{r, include = FALSE}
+Sys.setenv(YOURPACKAGE_KEY = secret_make_key())
+```
+
+Now (after you restart R), you'll be able to take advantage of a special `secret_encrypt()` and `secret_decrypt()` feature: the `key` argument can be the name of an environment variable, instead of the encryption key itself.
+In fact, this is most natural usage.
```
jennybc, line 280:
```suggestion
In fact, this is the most natural usage.
```

Gah, another PR to a PR I've made to my own PR.

vignettes/wrapping-apis.Rmd
jennybc: Words I have used for this topic in https://gargle.r-lib.org/articles/articles/managing-tokens-securely.html

> Decryption is not going to work. Either because the Suggested sodium package is not available or (much more likely) because the environment variable that represents the key is not available.
> This will be the case on CRAN, by definition, because there is no way to share an encrypted secret.
> This will be the case for external contributors, on their personal machines and when their GitHub pull requests are checked via CI services, such as GitHub Actions, Travis-CI, or AppVeyor.
> ...
> By calling the skipper at the start of tests that require auth, you arrange for your package to cope gracefully when the token cannot be decrypted, e.g., on CRAN and in pull requests.

PR #76, opened 2021-09-23
app -> package

maelle:
Another remark: Why not mention positive aspects of including an user agent, e.g. the API maintainers could value your feedback more if they see your wrapper is used a lot?


hadley:
Thanks! 

And because I don't think there's much evidence that this ever actually happens in practice.

PR #102, opened 2022-01-17
Explain device code behavior

m-mohr:
When using the device code flow I was a bit confused as it simply said "use code ...". Would you mind a description that is a bit longer?


m-mohr:
Uhm, why has this been closed?

hadley:
No idea. I didn't close it 😱

PR #103, opened 2022-01-17
Remove ambiguous interactive check

m-mohr:
Enables using device code in Jupyter environments (through IRKernel).


m-mohr:
image

The `is_interactive()` in the if is always true and not required if `interactive = TRUE` is present in the `oauth_flow_check`. Unfortunately, `IRKernel` for Jupyter is declared as non-interactive and as such the device code can't be used there. Would be nice if we could enable it for Jupyter envs, too.

m-mohr:
Workaround for us is now: `rlang::with_interactive(oauth_flow_device(...))`

Issue #92, opened 2021-10-07
Device Code Grant with PKCE


m-mohr:
Some providers (e.g. EGI check-in) support Device Code with PKCE (as in Auth Code with PKCE). This is useful in some circumstances and we'd like to use it instead of AuthCode with PKCE to avoid (1) spinning up a server in the background and (2) providing a client secret. Are there any plans to implement Device Code with PKCE? If not, would you be open to a PR?

flahn:
Please have a look at the PR  #109. I have implemented and tested the device code flow + PKCE against the [EGI service](https://www.egi.eu/services/check-in/) which we use in the [openEO projects](https://openeo.cloud/).

PR #109, opened 2022-02-25
Device Code + PKCE implementation

flahn:
A follow-up implementation for r-lib/httr2#92 which adds the option to use PKCE during the device code flow authentication. The implementation follows the existing one for the authentication code flow.

### DIFF
--- a/R/oauth-flow-device.R
+++ b/R/oauth-flow-device.R
@@ -75,14 +75,12 @@ oauth_flow_device <- function(client,
   # Google uses verification_url instead of verification_uri
   # verification_uri_complete is optional, it would ship the user
   # code in the uri https://datatracker.ietf.org/doc/html/rfc8628#section-3.2
-  if (is_interactive() && has_name(request, "verification_uri_complete")) {
+  url <- request$verification_uri_complete %||% request$verification_uri %||% request$verification_url
+
+  if (is_interactive()) {
     inform(glue("Use code {request$user_code}"))
     utils::browseURL(request$verification_uri_complete)
-  } else if (is_interactive() && has_name(request, "verification_uri")) {
-    inform(glue("Use code {request$user_code}"))
-    utils::browseURL(request$verification_uri)
   } else {
-    url <- request$verification_uri %||% request$verification_url
     inform(glue("Visit <{url}> and enter code {request$user_code}"))
   }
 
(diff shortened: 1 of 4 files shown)


R/oauth-flow-device.R
```diff
@@ -49,20 +54,33 @@ req_oauth_device <- function(req, client,
...
   # User interaction
   # https://datatracker.ietf.org/doc/html/rfc8628#section-3.3
   # Azure provides a message that we might want to print?
   # Google uses verification_url instead of verification_uri
+  # verification_uri_complete is optional, it would ship the user
+  # code in the uri https://datatracker.ietf.org/doc/html/rfc8628#section-3.2
   if (is_interactive() && has_name(request, "verification_uri_complete")) {
     inform(glue("Use code {request$user_code}"))
     utils::browseURL(request$verification_uri_complete)
+  } else if (is_interactive() && has_name(request, "verification_uri")) {
```
hadley, line 81: The problem with informing the user about the code and then immediately calling `browseURL` is that focus will switch to the browser before they have a chance to see the code in RStudio.

flahn: Yes, this is not ideal, indeed. However this was also the prior workflow. From a user perspective I find it easier to switch back from the browser to RStudio, where I know that I had made the call, rather than to look for the browser and potentially the tab, where I have to paste the device code into. This gets even more difficult if you have to C&P the URL also and the device code. I mean usually you will be asked in the device code flow to type in the device code. Then you start looking where to find the code and that would be where you had made the call.

Potential solutions to this:
* delaying the `browseURL` call with a timeout
* adding more informative text, how the user shall use the device in either the Rd file or directly printed to console

But this gets off-topic a bit. I mean, the intention of this PR is the enablement of PKCE in the device code flow and not so much the Human-Computer Interaction optimizations.

hadley: Oh I see what you mean. I think the construction of the `if` statement is throwing me off. Maybe it should be more like:

```R
url <- request$verification_uri_complete %||% request$verification_uri %||% request$verification_url

if (is_interactive()) {
  inform(glue("Use code {request$user_code}"))
  utils::browseURL(request$verification_uri_complete)
} else { 
  inform(glue("Visit <{url}> and enter code {request$user_code}"))
}
```

PR #121, opened 2022-04-01
Ensure expires_in is numeric

### LINKED ISSUE #118: Object type error
I'm trying the `req_oauth_password` flow (sorry no reprex) and I'm getting an error in `as.numeric(.date) + expires_in` within the `oauth_token()` function. I've debugged it and it seems like the `expires_in` variable is a string and also needs to be converted to a numeric.

jl5000:
Fix #118

### DIFF
--- a/R/oauth-token.R
+++ b/R/oauth-token.R
@@ -32,7 +32,7 @@ oauth_token <- function(
 
   if (!is.null(expires_in)) {
     # Store as unix time to avoid worrying about type coercions in cache
-    expires_at <- as.numeric(.date) + expires_in
+    expires_at <- as.numeric(.date) + as.numeric(expires_in)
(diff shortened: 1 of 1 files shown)


R/oauth-token.R
```diff
@@ -32,7 +32,7 @@ oauth_token <- function(
 
   if (!is.null(expires_in)) {
     # Store as unix time to avoid worrying about type coercions in cache
-    expires_at <- as.numeric(.date) + expires_in
+    expires_at <- as.numeric(.date) + as.numeric(expires_in)
```
hadley, line 35: Hmmm, I think the fix should probably be further upstream. And maybe `oauth_token()` should assert that `expires_in` is numeric?

PR #125, opened 2022-04-19
Make actual request rate equal to expected, fixes #101

### LINKED ISSUE #101: Understanding rate/throttle pool in `req_throttle`
I may just be doing something silly, but these timings look off to me. I am using the CRAN version on Ubuntu 20.04 and R 4.1.2.

```
library(dplyr)
library(httr2)

get_throttled <- function(url, rate) {
  message(Sys.time())
  request(url) %>%
    req_throttle(rate) %>%
    req_method("get") %>% 
    req_perform()
}

urls <- rep("https://httpbin.org/status/200", 60)

system.time(x <- lapply(urls, get_throttled, rate = 1))

#  user  system elapsed 
# 0.666   0.125  34.302

system.time(x <- lapply(urls, get_throttled, rate = 30/60))

#  user  system elapsed 
# 0.826   0.199  64.362

system.time(x <- lapply(urls, get_throttled, rate = 10/60))

#  user  system elapsed 
# 1.344   0.646 183.221
```

jchrom:
Fixes the problem with `throttle_delay()` where the actual time requests take is half of what you would expect.

### What I've done:

- moved `sys_sleep()` inside `throttle_delay()` so that `throttle_touch()` can happen after it and still find the correct realm
- removed the original `sys_sleep()`
- added `sys_sleep()` where appropriate to preserve the expected behavior on backoff, reauth and retry after

### Tests

The original test in _test-req-throttle.R_ passes as is, because the value of `wait` is still returned (now only useful for testing).

I also tested manually using httpbin to ensure times are now as expected.

I experimented with including tests where I was issuing requests via a fake app, so I could measure the elapsed time. This worked, but when set to very high rates, even a slight discrepancy in wait time would make the test fail on occasion, unless a high tolerance was used, which then kind of defeats the test. Lower request rates were more reliable, but then the test takes a long time to execute. It could be skipped on CRAN though.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httr2 (development version)
 
+* `req_throttle()` correctly sets throttle rate (@jchrom, #101).
+
 * `req_error()` can now correct force successful HTTP statuses to fail (#98).
 
 * `req_headers()` will now override `Content-Type` set by `req_body_*()` (#116)
--- a/R/req-perform.R
+++ b/R/req-perform.R
@@ -70,13 +70,13 @@ req_perform <- function(
 
   n <- 0
   tries <- 0
-  delay <- throttle_delay(req)
   reauth <- FALSE # only ever re-authenticate once
+
+  throttle_delay(req)
 
   while(tries < max_tries && Sys.time() < deadline) {
     n <- n + 1
 
-    sys_sleep(delay)
     resp <- tryCatch(
       req_perform1(req, path = path, handle = handle),
       error = function(err) {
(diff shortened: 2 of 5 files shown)


hadley:
Thanks for working on this! For the test, did you consider something like this?

``` r
library(httr2)
req <- httr2:::request_test() %>% req_throttle(1000 / 60)
times <- replicate(20, system.time(req_perform(req)))
mean(times[3, ], trim = 0.10)
#> [1] 0.0675625
```

<sup>Created on 2022-04-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I think you're better of recording the individual times because then you can throw out a few that are particularly slow.

jchrom:
That makes a lot of sense, let me update

jchrom:
Finicky!

hadley (approved):
Thanks! Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
I'll finish this off 😄

PR #128, opened 2022-04-22
Keep duplicates in `modify_list()`, fixes #97, fixes #107

### LINKED ISSUE #97: Stacking queries in req_url_query()
In some APIs, such as for [Jive](https://developers.jivesoftware.com/api/v3/cloud/rest/index.html), you can **stack** query parameters with the same name and differentiate them in the values, however the default behaviour of `httr2` seems to be to overwrite them if they have the same name, e.g.:

`httr2::req_url_query(req, filter = "name(Hadley)", filter = "company(rstudio)")`

Would it be worth adding a parameter to switch off this behaviour?

### LINKED ISSUE #107: `req_body_form()` silently drops duplicate names
In order to POST a form on a website where all the checkboxes have identical names, I need to pass a list with duplicate names. But, the duplicates are silently dropped.

The culprit seems to be `.x[names(dots)] <- dots` in [modify_list()](https://github.com/r-lib/httr2/blob/87011c0ff31019409c4a5a700b041f279f054361/R/utils.R#L31).

It could be replaced with something like

```
.x[names(.x) %in% names(dots)] <- NULL
.x <- c(.x, dots)
```

which would have the advantage of replacing entire homonym groups, and would pass the current tests for `modify_list()`.

It wouldn't help with the headaches in #95 ~and # 88~ though (I am not sure about the # 88).

For a more systematic solution, how about an additional argument `replace = TRUE`, or a dedicated `req_body_form_modify()` so that we can decide whether the body gets replaced instead of combined?

- `req_body_*`: replaces entire body, avoiding the `modify_list()` call
- `req_body_*_modify`: combines lists

jchrom:
# The issue

When `...` in `modify_list()` includes duplicates, they get silently dropped. This is inconvenient mainly in `req_body_*` and `req_url_query()`. Note e.g. the WHATWG URL Standard seems to allow for non-unique query parameters (see this [nice SO answer](https://stackoverflow.com/a/70264621) with references).

# Change in default behavior

```
x <- list(m = 1)
httr2:::modify_list(x, n = 1, n = 2)
```

Result before change: 

`list(m = 1, n = 2)`

Result after change:

`list(m = 1, n = 1, n = 2)`

# Tests

Added a test for preserving duplicates. All previous tests pass locally as well.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -47,10 +47,14 @@ replace2 <- function(x, y) {
     xi <- which(names(x) == nm)
     yi <- which(names(y) == nm)
 
+    # Replace all occurrences of `nm` in `x` with the same number of occurrences
+    # from `y`.
     x[xi[seq_along(xi) <= length(yi)]] <- y[yi[seq_along(yi) <= length(xi)]]
 
+    # If there is a leftover on the part of `x`, it is dropped.
     x[xi[seq_along(xi) > length(yi)]] <- NULL
 
+    # If leftover happens on the part of `y`, it is appended.
     x <- c(x, y[yi[seq_along(yi) > length(xi)]])
   }
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -21,15 +21,16 @@ bullets_with_header <- function(header, x) {
   cli::cli_li(paste0("{.field ", names(x), "}: ", vals))
 }
 
-modify_list <- function(.x, ...) {
+modify_list <- function(.x, ..., .compact = TRUE) {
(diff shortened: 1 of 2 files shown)


jchrom:
Fails on the unrelated throttle test.

hadley:
I think `modify_list()` is used in too many other places to be the right thing to modify. Let me think about this one for a bit.

hadley:
No, I think you're right, but there's an easier way to do this. PR incoming.

PR #135, opened 2022-04-30
Allow appending path in `req_template()`, fixes #133

### LINKED ISSUE #133: Make `req_template()` append the path, instead of replace
It is somewhat inconvenient that when modifying an existing request, the path gets overwritten wholly. Using `req_url_path_append()` instead of `req_url_path()` would be more convenient.

Could we squeeze this in for 0.2, if it makes sense?

jchrom:
## The issue

`req_template()` overwrites existing path in `req`

## Changes

- added an additional `.append = FALSE` argument to allow appending the path instead, leaving default behavior unchanged
- added a test to `test-req-template.R`
- added an example to function docs, and a new section

In the end I went with adding an argument, but I am happy to change if you think there should only be a single way. I tried to be clear in the docs in a new section to `req_template()` but suggestions are welcome.

### DIFF
--- a/R/req-template.R
+++ b/R/req-template.R
@@ -3,8 +3,16 @@
 #' @description
 #' Many APIs document their methods with a lightweight template mechanism
 #' that looks like `GET /user/{user}` or `POST /organisation/:org`. This
-#' function makes it easy to copy and paste this snippets and retrieve template
-#' variables either from function arguments or the current environment:
+#' function makes it easy to copy and paste these snippets and retrieve template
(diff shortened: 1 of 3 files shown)


jchrom:
Ahh, these throttle tests. They are super finicky. Should we try and do anything about that?

jchrom:
To be honest, now that I look at the solution, I am no longer sure this is preferable to some kind of a prefix argument.

hadley:
I think you should just always append.

jchrom:
Ok, incoming.

PR #136, opened 2022-05-02
Make `req_template()` append path instead of replace, fixes #133

### LINKED ISSUE #133: Make `req_template()` append the path, instead of replace

jchrom:
I added a test and updated the docs, hopefully clearly enough. I have also updated `request_test()` because `req_url_path_append()` respects double slash. Which, as stated in `test-req-url.R`, is intentional, if a bit confusing to me.


jchrom:
Thanks, I forgot to include the issue number in the commit message, and then tried to cheat by editing the title. I wonder if the `fixes` keyword will work now that I added it to the subsequent commit?

Anyway I added the bullet under the `# httr2 (development version)` heading.

R/test.R
```diff
@@ -7,7 +7,7 @@ request_test <- function(template = "/get", ...) {
   }
 
   req <- request(the$test_app$url())
-  req <- req_template(req, template, ..., .env = caller_env())
+  req <- req_template(req, sub("^/", "", template), ..., .env = caller_env())
```
hadley, line 10: This doesn't seem right to me — should this logic belong in `req_url_path_append()`?

I think I'd expect:

```
req1 <- request("http://test.com/x") %>% req_template("PATCH /y")
req2 <- request("http://test.com/x/") %>% req_template("PATCH /y")
expect_equal(req1$url, req2$url)
```

jchrom: That's what I would expect! But then I noticed in `test-req-url.R` (line 19):

```
req2 <- request("http://example.com/")

# specifically requested to add in both
  expect_equal(req_url_path_append(req2, "/index.html")$url, "http://example.com//index.html")
```

I don't know why but it is on purpose. So I left it as it was.

hadley: I think we should say that `req_url_path_append()` will never make a `//` — if you do need it for some reason, you can use `req_url_path()`.

jchrom: Ok. What's the best way to handle this - a new PR?

tests/testthat/test-req-url.R
```diff
@@ -15,8 +15,6 @@ test_that("automatically adds /", {
   expect_equal(req_url_path_append(req1, "index.html")$url, "http://example.com/index.html")
   expect_equal(req_url_path_append(req1, "/index.html")$url, "http://example.com/index.html")
   expect_equal(req_url_path_append(req2, "index.html")$url, "http://example.com/index.html")
-  # specifically requested to add in both
-  expect_equal(req_url_path_append(req2, "/index.html")$url, "http://example.com//index.html")
```
hadley, line 19: Can you keep this test but update the expected value?

R/req-url.R
```diff
@@ -69,10 +69,10 @@ req_url_path_append <- function(req, ...) {
 
   url <- url_parse(req$url)
 
-  if (!grepl("^/", path) && (is.null(url$path) || !grepl("/$", url$path))) {
+  if (!grepl("^/", path)) {
     path <- paste0("/", path)
   }
 
-  url$path <- paste0(url$path, path)
+  url$path <- paste0(sub("/$", "", url$path), path)
```
hadley, line 76: I'd move the `sub()` up to the block above so all the `/` handling stuff is in one place.

hadley (commented):
Looks good; this will need a news bullet and an update to the docs for `req_url_append_path()`. Maybe something like: `req_url_append_path()` always appends a directory or file name, i.e. it always ensures that there's exactly one `/` in between the old path and the new path.

tests/testthat/test-req-url.R
jchrom: Done, [14b7aa0](https://github.com/r-lib/httr2/pull/136/commits/14b7aa00ea0ddb1db33b83c85faf0c3df7d1c72c) has that in it.

R/req-url.R
jchrom: Hmm, it's getting late around here, so I might not be getting exactly what you mean. Something like

```
url <- url_parse(req$url)

url$path <- sub("/$", "", url$path)

if (!grepl("^/", path)) {
  path <- paste0("/", path)
}

url$path <- paste0(url$path, path)
```
?

jchrom:
> req_url_append_path() always appends a directory or file name, i.e. it always ensures that there's exactly one / in between the old path and the new path.

To me this sounds like any number of slashes above one gets removed, but in reality:

``` r
library(httr2)
request("https://httpbin.org/hi") %>% req_url_path_append("///bye")
#> <httr2_request>
#> GET https://httpbin.org/hi///bye
#> Body: empty
```

<sup>Created on 2022-05-05 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Should we update the regex so that it makes sure multiple slashes get dropped? Or make the docs say that "no extra slashes are introduced" rather than saying what the final state would be?

R/req-url.R
hadley: Yeah, except I'd do:

```R
url$path <- sub("/$", "", url$path)
if (!grepl("^/", path)) {
  path <- paste0("/", path)
}
```

to make it more clear that it's a "paragraph" of related code

hadley:
Hmmm good question, maybe say that it will insert or remove a slash if needed? I'm not sure how to phrase it to minimise confusion.

R/req-template.R
```diff
@@ -3,8 +3,12 @@
...
 #' Many APIs document their methods with a lightweight template mechanism
 #' that looks like `GET /user/{user}` or `POST /organisation/:org`. This
-#' function makes it easy to copy and paste this snippets and retrieve template
-#' variables either from function arguments or the current environment:
+#' function makes it easy to copy and paste such snippets and retrieve template
+#' variables either from function arguments or the current environment.
+#'
+#' `req_template()` will append to the existing path so that you can set a
+#' base url in the initial [request()]. This means that you'll generally want
+#' to avoid `req_template()` calls to the same request.
```
hadley, line 11:
```suggestion
#' to avoid multiple `req_template()` calls on the same request.
```

hadley:
I'm going to finish this off and merge it, just because I need to submit httr2 to avoid it getting pulled from CRAN. Happy to continue to iterate on the documentation wording.

PR #138, opened 2022-05-03
Added "f" option to open in req_stream (fix #137)

### LINKED ISSUE #137: Unable to handle req_stream errors due to stream open throwing an error
I'm writing an API wrapper that POSTs a command to the server via req_stream to select which data to return. If the posted body is malformed, the server responds with error code 400 causing the open command within req_stream to throw an error. The error prevents the subsequent reading of the error contents by the callback and prevents the subsequent creation of a new_response return value containing the status code.

The R curl_fetch_stream function does not suffer this same limitation because it opens the stream using "rbf" mode instead of "rb" mode. According to the curl's internal code comment, "# 'f' means: do not error for status code".

I request that req_stream be modified such that

stream <- curl::curl(req$url, handle = handle)
open(stream, "rb")

becomes

stream <- curl::curl(req$url, handle = handle)
open(stream, "rbf")

My local testing shows that this change prevents R from throwing an error, allows subsequent parsing of the stream contents, and correctly creates the new_response code with the observed status code.

DMerch:


hadley (approved):
Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

Also for future reference, `fix #137` only works in the description of the PR, not the title.

DMerch:
I added the note.  Sorry I'm just now learning GitHub development procedures.

hadley:
Thanks!

PR #145, opened 2022-05-17
Bug fix in `auth_flags`

jl5000:
The code is erroring for me at this point and I'm pretty sure it's because the `idx` variable isn't being used.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httr2 (development version)
 
+* `req_proxy()` now uses the appropriate authentication option (@jl5000).
+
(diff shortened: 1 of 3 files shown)


hadley:
Ooops. Would you mind adding a small test too?

hadley:
Sorry I forgot to ask this earlier, but can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

hadley:
Thanks!

PR #164, opened 2022-08-03
More robust `parse_media()`

### LINKED ISSUE #163: Error when printing response where content type is empty string
I encountered a response I cannot print. It boiled down to the response having the empty string as content type.

``` r
httr2::response(headers = list(`Content-Type` = ""))
#> <httr2_response>
#> GET https://example.com
#> Status: 200 OK
#> Error in pieces[[1]]: subscript out of bounds
```

<sup>Created on 2022-08-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

mgirlich:
Fixes #163.
Similar issues could happen in `parse_www_authenticate()` and `parse_link()`. I'm not sure whether that could really happen but they use the same pattern.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httr2 (development version)
 
+* Can now print responses where content type is the empty string (@mgirlich, #163).
+
 * `curl_translate()` can now handle curl copied from Chrome developer tools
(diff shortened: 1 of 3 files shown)


hadley:
Thanks!

PR #168, opened 2022-09-13
Fix oauth refresh

### LINKED ISSUE #166: Getting`Error: $ operator is invalid for atomic vectors` with oauth flow token refresh
I am getting an error whenever my token expires and there is an attempt to auto-refresh the token. This is the function I am calling, with some information redacted for privacy purposes. The function takes a client that I instantiate with the `httr2::oauth_client()` function.

```
api_request_function <- function(client) {
  httr2::request("[MY_API_URL]") |>
    httr2::req_url_path_append("[PATH_ONE]") |>
    httr2::req_url_path_append("[PATH_TWO]") |>
    httr2::req_oauth_client_credentials(
      client,
      scope = paste(
        "[SCOPE_ONE]",
        "[SCOPE_TWO]"
      ),
      token_params = list(
        "[PARAM_ONE]" = "[XXX]",
        "[PARAM_TWO]" = "[XXX]",
        "[PARAM_THREE]" = "[XXX]"
      )
    ) |>
    httr2::req_retry(max_tries = 3) |>
    httr2::req_perform() |>
    httr2::resp_body_json()
}
```

The function works perfectly fine when the token hasn't expired, but once it expires I get the following error and traceback:

```
Error: $ operator is invalid for atomic vectors

11. is_string(url)
10. new_request(base_url)
9. request(client$token_url)
8. oauth_client_get_token(client, grant_type = "refresh_token", refresh_token = refresh_token, scope = scope, !!!token_params)
7. token_refresh(flow$params$client, token)
6. auth_oauth_sign(req)
5. httr2::req_perform(httr2::req_retry(req_oauth_client_mosaic(httr2::req_url_path_append(httr2::req_url_path_append(httr2::request("[MY_API_URL]"), "[PATH_ONE]"), "[PATH_TWO]"), client = client), max_tries = 3))
4. is_response(req)
3. check_response(resp)
2. httr2::resp_body_json(httr2::req_perform(httr2::req_retry(req_oauth_client_mosaic(httr2::req_url_path_append(httr2::req_url_path_append(httr2::request("[MY_API_URL]"), "[PATH_ONE]"), "[PATH_TWO]"), client = client), max_tries = 3)))
1. api_request_function(client)
```

After getting this error, if I try running the function again it works perfectly fine until the token expires again, and then I get the same error. It seems like there might be an issue with refreshing the token? I am using version 0.2.1 of the `httr2` package. Happy to provide more information if needed!

jennybc:
This fixes #166. ~Sort of.~ *I believe I've solved the multiple refresh problem also now.*

In my hands, I can only use a valid refresh token twice:

First, in the initial call that triggers the oauth dance in the browser. (With `cache_disk = TRUE`, the token that is written to disk includes both the refresh and access token.)

Then, with this PR, an initial refresh works, i.e. a new access token is obtained.

**But** after this first refresh, it looks like the token cached on disk, no longer includes the refresh token? In which case the next (3rd) attempt leads to a full oauth dance, when it should not. I'm still verifying and investigating this claim about the token on disk.

It's a slow process because to truly test things, you really do have to let access tokens expire. I'm testing against Google, where access tokens last for 60 minutes.

We could merge this, because it's progress. Or wait until I clear up the other matter.

### DIFF
--- a/R/oauth-token.R
+++ b/R/oauth-token.R
@@ -78,10 +78,13 @@ token_has_expired <- function(token, delay = 5) {
 }
 
 token_refresh <- function(client, refresh_token, scope = NULL, token_params = list()) {
-  oauth_client_get_token(client,
+  out <- oauth_client_get_token(
+    client,
     grant_type = "refresh_token",
     refresh_token = refresh_token,
     scope = scope,
     !!!token_params
   )
+  out$refresh_token <- out$refresh_token %||% refresh_token
+  out
 }
--- a/R/oauth.R
+++ b/R/oauth.R
@@ -46,7 +46,7 @@ auth_oauth_sign <- function(req, reauth = FALSE) {
       if (is.null(token$refresh_token)) {
         token <- exec(flow, !!!flow_params)
       } else {
-        token <- token_refresh(flow$params$client, token)
+        token <- token_refresh(flow_params$client, token)
       }
     }
   }
--- a/R/oauth.R
+++ b/R/oauth.R
@@ -46,7 +46,7 @@ auth_oauth_sign <- function(req, reauth = FALSE) {
       if (is.null(token$refresh_token)) {
         token <- exec(flow, !!!flow_params)
       } else {
-        token <- token_refresh(flow_params$client, token)
+        token <- token_refresh(flow_params$client, token$refresh_token)
       }
     }
   }
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httr2 (development version)
 
+* OAuth tokens can now be refreshed. One, two, or even more times! (@jennybc, #166)
+
 * Can now print responses where content type is the empty string (@mgirlich, #163).
 
 * `curl_translate()` can now handle curl copied from Chrome developer tools
--- a/R/oauth-flow-refresh.R
+++ b/R/oauth-flow-refresh.R
@@ -2,7 +2,7 @@
 #'
 #' @description
 #' This uses [oauth_flow_refresh()] to generate an access token, which is
-#' then used to authentication the request with [req_auth_bearer_token()].
+#' then used to authenticate the request with [req_auth_bearer_token()].
 #' This is primarily useful for testing: you can manually execute another OAuth
 #' flow (e.g. by calling [oauth_flow_auth_code()] or [oauth_flow_device()]),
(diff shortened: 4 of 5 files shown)


R/oauth.R
```diff
@@ -46,7 +46,7 @@ auth_oauth_sign <- function(req, reauth = FALSE) {
       if (is.null(token$refresh_token)) {
         token <- exec(flow, !!!flow_params)
       } else {
-        token <- token_refresh(flow$params$client, token)
+        token <- token_refresh(flow_params$client, token$refresh_token)
```
jennybc, line 49: The switch from `flow$params$client` to `flow_params$client` is the immediate fix to what's reported in #166. Once you fix that, a second problem is revealed now that the refresh process can progress farther along:

```
# Error in `query_build()` at httr2/R/req-body.R:219:4:
# ! Query parameters must be length 1 atomic vectors.
# • Problems: refresh_token
```

And that explains the switch from `token` to `token$refresh_token`.

hadley (commented):
I re-read the code and this change looks correct

R/oauth-token.R
```diff
@@ -78,10 +78,13 @@ token_has_expired <- function(token, delay = 5) {
...
 token_refresh <- function(client, refresh_token, scope = NULL, token_params = list()) {
-  oauth_client_get_token(client,
+  out <- oauth_client_get_token(
+    client,
     grant_type = "refresh_token",
     refresh_token = refresh_token,
     scope = scope,
     !!!token_params
   )
+  out$refresh_token <- out$refresh_token %||% refresh_token
```
jennybc, line 88: My understanding of the overall design is still tenuous enough that I'm not 100% sure of where to do this, but I think this is the right place.

For Google, the response for a token refresh looks like this:

```
{
  "access_token": "1/fFAGRNJru1FTz70BzhT3Zg",
  "expires_in": 3920,
  "scope": "https://www.googleapis.com/auth/drive.metadata.readonly",
  "token_type": "Bearer"
}
```

I confirmed this personally and it's also documented at <https://developers.google.com/identity/protocols/oauth2/native-app#offline>.

The main thing to notice is the LACK of the refresh token. So I think we need to proactively preserve it.

jennybc:
@Hadley I think I found and fixed the problem preventing 2nd and subsequent refreshes.

hadley:
Thanks for all your work on this!

PR #169, opened 2022-09-14
Rename env var

jennybc:
I don't think this needs a NEWS bullet, but let me know if you want one.


hadley (approved):
Probably worth a news bullet just in case

PR #171, opened 2022-09-19
removed oauth_flow_check() in oauth_flow_device

### LINKED ISSUE #170: Problems with Device Code + PKCE authentication on remote machines
Hi all,

I had made some contributions to the device code authentication with PKCE in the past and we have used it as our main authentication procedure, which works under normal circumstances (local client) perfectly fine.

Recently, we have setup up a Jupyter notebook with an R Kernel on a remote server. That's where the problem begins. The Jupyter  environment is regarded as an interactive session `rlang::is_interactive()` and httr2 tries to open up the internet browser that is defined in Rs global environment. Unfortunetly the user never gets to see the browser, because it gets executed on the remote machine. And usually you would expect that the URL and the code is printed to console in order to open the browser on your local machine, as it is already defined [here](https://github.com/r-lib/httr2/blob/487c23e203e38539198432c025fc16b48bf2d498/R/oauth-flow-device.R#L85).

Our workaround for this in the Jupyter notebook would be to run the authentication in a temporary non-interactive state with  `rlang::with_interactive({...}, value=FALSE)`. This way we keep the general interactiveness of Jupyter and disable it once during authentication. So far, so good.

Now, there is still a big issue with this approach. [This line](https://github.com/r-lib/httr2/blob/487c23e203e38539198432c025fc16b48bf2d498/R/oauth-flow-device.R#L61) is blocking this non-interactive state. As I understand, the intention for `oauth_flow_check(...,interactive=TRUE)` is a legacy thing, that is was assumed that the user needs to be actively involved by direct interaction with the browser. To cut it short, this line throws an error, because this "interactive state" is required and prevents the printing of the URL and the code from ever being called. 

My question is whether the `oauth_flow_check` is really required here or the default value for `interactive` could be changed to `rlang::is_interactive()`?

Best regards,
Florian

flahn:
- fixes #170
- checking interactiveness at that point prevented the code block about printing url and device code for manual browser interaction from being ever executed

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httr2 (development version)
 
+* disabled check for interactive session `oauth_flow_check(..., interactive)` on device code flow to print URL and device code as intended (@flahn, #170)
+
 * The environment variable `HTTR2_REFRESH_TOKEN` replaces the previous `HTTR_REFRESH_TOKEN` (@jennybc, #169).
 
 * OAuth tokens can now be refreshed. One, two, or even more times! (@jennybc, #166)
--- a/R/oauth-flow-device.R
+++ b/R/oauth-flow-device.R
@@ -58,7 +58,6 @@ oauth_flow_device <- function(client,
                               scope = NULL,
                               auth_params = list(),
                               token_params = list()) {
-  oauth_flow_check("device", client, interactive = TRUE)
 
   if (pkce) {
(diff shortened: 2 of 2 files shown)


hadley:
Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

Issue #165, opened 2022-08-21
Can't provide custom `redirect_uri` in `oauth_auth_code()`


kelvindso:
Hi team,

Thanks for providing this wonderful HTTP library.

When using httr2 to access the Twitter API v2, I ran the following commands in a R console:
```R
client <- oauth_client(
  id = "XYZ",
  secret = "123",
  token_url = "https://api.twitter.com/2/oauth2/token",
  name = "oauth-test"
)
req <- request("https://api.twitter.com/2/tweets/1560771328814108673") %>%
  req_oauth_auth_code(client, auth_url = "https://twitter.com/i/oauth2/authorize", scope = "tweet.read", auth_params = list(state = "state", redirect_uri = "https://rest.com/callback"))
req %>% req_perform()
```

It triggered the browser to access
```
https://twitter.com/i/oauth2/authorize?response_type=code&client_id=XYZ&redirect_uri=http%3A%2F%2Flocalhost%3A46682%2F&scope=tweet.read&state=VP57MO24gVEvA1mh6Rgdsn74_qca90AZ1UsyqbGTmFQ&state=state&redirect_uri=https%3A%2F%2Frest.com%2Fcallback&code_challenge=AvF3Y5aEHVuhopICpVBnwYVh6D2D5LkrufAy7_1RKcY&code_challenge_method=S256
```
But unfortuately the redirect_uri is provided twice in the URL query parameters
- `redirect_uri=http%3A%2F%2Flocalhost%3A46682%2F`
- `redirect_uri=https%3A%2F%2Frest.com%2Fcallback`

Twitter API backend seems to only allow https for redirect URI and the redirect_url value must be the same as the one registered in Twitter developer portal (e.g. https://rest.com/callback)

Passing `host_name` to `oauth_flow_auth_code` didn't work either. 

Anyone knows what's the correct way to pass the redirect_url?

Thank you.

Kelvin

PR #173, opened 2022-10-09
Add redirect_uri argument to oauth_flow_auth_code method

wing328:
Add redirect_uri argument to oauth_flow_auth_code method instead of creating the the redirect URL from host_name and port.

This should fix https://github.com/r-lib/httr2/issues/165.


hadley:
Thanks for working on this but I think it needs a bit more of a broader look:  if we're going to provide this parameter, then  we need to rethink the `host_name` and `host_ip` arguments too. And then `port` also seems to have an inconsistent name (no prefix).

PR #177, opened 2022-10-20
`req_url_path_*()` can handle empty `...`

mgirlich:
Before this PR they errored with `argument is of length zero`. Now `req_url_path_append(req, NULL)` works.

### DIFF
--- a/R/req-url.R
+++ b/R/req-url.R
@@ -53,8 +53,7 @@ req_url_query <- function(.req, ...) {
 #' @rdname req_url
 req_url_path <- function(req, ...) {
   check_request(req)
-  path <- paste(..., sep = "/")
-  path <- path_add_slash(path)
+  path <- dots_to_path(...)
 
   req_url(req, url_modify(req$url, path = path))
 }
@@ -63,19 +62,18 @@ req_url_path <- function(req, ...) {
 #' @rdname req_url
 req_url_path_append <- function(req, ...) {
   check_request(req)
-  path <- paste(..., sep = "/")
-  path <- path_add_slash(path)
+  path <- dots_to_path(...)
 
   url <- url_parse(req$url)
-
   url$path <- paste0(sub("/$", "", url$path), path)
 
   req_url(req, url_build(url))
 }
 
-path_add_slash <- function(path) {
+dots_to_path <- function(...) {
+  path <- paste(c(...), sep = "/", collapse = "/")
   # Ensure we don't add duplicate /s
-  if (!is_empty(path) && !grepl("^/", path)) {
+  if (path != "" && !grepl("^/", path)) {
     path <- paste0("/", path)
   }
 
--- a/R/req-url.R
+++ b/R/req-url.R
@@ -71,7 +71,7 @@ req_url_path_append <- function(req, ...) {
 }
 
 dots_to_path <- function(...) {
-  path <- paste(c(...), sep = "/", collapse = "/")
+  path <- paste(c(...), collapse = "/")
   # Ensure we don't add duplicate /s
   if (path != "" && !grepl("^/", path)) {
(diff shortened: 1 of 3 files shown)


R/req-url.R
```diff
@@ -67,14 +64,20 @@ req_url_path <- function(req, ...) {
...
+  path <- path_add_slash(path)
 
   url <- url_parse(req$url)
 
+  url$path <- paste0(sub("/$", "", url$path), path)
+
+  req_url(req, url_build(url))
+}
+
+path_add_slash <- function(path) {
```
hadley, line 76: I think we should make this `path_add_leading_slash()` just to be very specific.

R/req-url.R
```diff
@@ -67,14 +64,20 @@ req_url_path <- function(req, ...) {
...
   # Ensure we don't add duplicate /s
-  if (!grepl("^/", path)) {
+  if (!is_empty(path) && !grepl("^/", path)) {
```
hadley, line 78: This adds special handling for the length-0 case, which makes me think we should also consider the length >1 case, which suggests that maybe instead of 

```R
  path <- paste(..., sep = "/")
```

we need

```R
  path <- paste(..., sep = "/", collapse = "/")
```

which will ensure that `path` is always a length-1 character vector and simplify this code path.

mgirlich: Hmm, then this would work

``` r
req <- request("http://example.com/x")
req_url_path_append(req, c("a", "b"), "c")
#> <httr2_request>
#> GET http://example.com/x/a/c/b/c
#> Body: empty
```

<sup>Created on 2022-10-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

which feels a bit weird to me. The alternative would be to check the size of path (empty or length 1). What do you prefer?

hadley: It's nice to allow this though:

```R
pieces <- c("a", "b", "c")
req <- request("http://example.com/x")
req_url_path_append(req, pieces)
```

mgirlich: Or support splicing?

```r
pieces <- c("a", "b", "c")
req <- request("http://example.com/x")
req_url_path_append(req, !!!pieces)
```

hadley: IMO splicing seems overkill when the input is an atomic vector because `c(c(x), c(y))` is the same as `c(x, y)`

R/req-url.R
```diff
@@ -66,15 +62,20 @@ req_url_path <- function(req, ...) {
...
+  url$path <- paste0(sub("/$", "", url$path), path)
 
+  req_url(req, url_build(url))
+}
+
+dots_to_path <- function(...) {
+  path <- paste(c(...), sep = "/", collapse = "/")
   # Ensure we don't add duplicate /s
-  if (!grepl("^/", path)) {
+  if (path != "" && !grepl("^/", path)) {
```
mgirlich, line 76: This is necessary b/c `paste(collapse = "/")` returns `""` instead of `character()`.

R/req-url.R
```diff
@@ -66,15 +62,20 @@ req_url_path <- function(req, ...) {
...
+  path <- dots_to_path(...)
 
   url <- url_parse(req$url)
+  url$path <- paste0(sub("/$", "", url$path), path)
 
+  req_url(req, url_build(url))
+}
+
+dots_to_path <- function(...) {
+  path <- paste(c(...), sep = "/", collapse = "/")
```
mgirlich, line 74: The `c(...)` is needed because

```r
paste(c("a", "b"), NULL, collapse = "/", sep = "/")
#> "a//b/"
```

hadley: I think you can then drop the `sep`?

mgirlich:
@hadley Is there anything missing to get this merged?

PR #190, opened 2023-01-09
Allow custom content type for JSON body

### LINKED ISSUE #189: Allow custom content type for JSON body
The content type is simply overwritten when using `req_body_json()`

``` r
library(httr2)

request("http://httpbin.org/404") |> 
  req_body_json(data.frame(x = 1)) |> 
  req_headers(`content-type` = "blub") |> 
  req_dry_run()
#> POST /404 HTTP/1.1
#> Host: httpbin.org
#> User-Agent: httr2/0.2.2.9000 r-curl/4.3.3 libcurl/7.85.0
#> Accept: */*
#> Accept-Encoding: deflate, gzip
#> Content-Type: application/json
#> Content-Length: 9
#> 
#> [{"x":1}]
```

<sup>Created on 2023-01-09 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### LINKED ISSUE #284: check_content_type() should correctly error when response has no content_type but suffix is checked.
This is issue is the same error as https://github.com/r-lib/httr2/issues/277 but in its minimal form. 

I stumbled upon this error while trying to do some mocking in testing with `httr2::response()`. 

````r
> httr2::resp_body_json(httr2::response())
Error in `if (!is.null(suffix) && endsWith(content_type, suffix)) ...`:
! missing value where TRUE/FALSE needed
> rlang::last_trace()
<error/rlang_error>
Error in `if (!is.null(suffix) && endsWith(content_type, suffix)) ...`:
! missing value where TRUE/FALSE needed
---
Backtrace:
    ▆
 1. └─httr2::resp_body_json(httr2::response())
 2.   └─httr2:::check_content_type(...)
```` 

I don't think the error is very clear to understand is the call is really allowed or not. 

Simpler reprex:
```r
httr2:::check_content_type(httr2::response(), "application/json", suffix = "+json")
```

In there, 
https://github.com/r-lib/httr2/blob/7962bcd9aacd1cc71ee34c5e4fd3c0b8d172d670/R/resp-body.R#L130-L138
`content_type` can be `NA_character_`
https://github.com/r-lib/httr2/blob/d789827ce2795edbfac92ddaf3c3df5e33e85df4/R/resp-headers.R#L101-L107

and this will make `endsWith()` returns a NA
``` r
endsWith(NA_character_, "+json")
#> [1] NA
```

which is not good in the `if()` call and with `&&`

#277 was fixed by #278 but regarding the specific cache usage. 

I believe here the error should be clearer. Maybe `is.na(content_types)` should be checked also when `!is.null(suffix)` to handle this specific `content_types` value, and throw useful error.

mgirlich:
Fixes #189. Fixes #284.

This PR doesn't support parameters in the media type. They could be added but I never encountered them and I'm not sure the extra complexity is worth it.
For `req_body_form()` it only allows `application/x-www-form-urlencoded`. I'm happy to add support for the more general form similar to JSON but I'm not sure it's worth it.


R/req-body.R
```diff
@@ -226,6 +226,39 @@ req_body_apply <- function(req) {
   req
 }
 
+check_json_content_type <- function(content_type, call = caller_env()) {
```
hadley, line 229: I wonder if it's worth slightly extending `check_content_type()` for this? (i.e. make it work with both requests and responses.

R/req-body.R
```diff
@@ -226,6 +226,39 @@ req_body_apply <- function(req) {
...
+  check_string(content_type)
+
+  if (is.null(content_type) || identical(content_type, "application/json")) {
+    return("application/json")
+  }
+
+  if (!startsWith(content_type, "application/")) {
```
hadley, line 236: Can you add a reference for this in a comment?

R/req-body.R
mgirlich: Yes, seems worth it.

tests/testthat/_snaps/resp-body.md
```diff
@@ -11,17 +11,15 @@
     Code
       request_test("/xml") %>% req_perform() %>% resp_body_json()
     Condition
-      Error in `check_content_type()`:
+      Error in `resp_body_json()`:
       ! Unexpected content type 'application/xml'
       Expecting 'application/json'
-      Or suffix '+json'
```
mgirlich, line 17: Not sure the information about the suffix are worth to implement.

tests/testthat/_snaps/req-body.md
```diff
@@ -6,13 +6,15 @@
     Output
       <error/rlang_error>
       Error in `req_body_apply()`:
-      ! Content type must end with "json" for a JSON body.
+      ! Unexpected content type 'application/ld+json2'
+      * Expecting 'application/json'
```
hadley, line 10: The info about the suffix would be nice here

tests/testthat/_snaps/resp-body.md
```diff
@@ -11,17 +11,17 @@
     Code
       request_test("/xml") %>% req_perform() %>% resp_body_json()
     Condition
-      Error in `check_content_type()`:
+      Error in `resp_body_json()`:
       ! Unexpected content type 'application/xml'
-      Expecting 'application/json'
-      Or suffix '+json'
+      i Expecting 'application/json' or 'application/<subtype>+json'
```
mgirlich, line 16: It is a bit more verbose than "or suffix +json" but might be clearer. What do you think?

hadley: I like it!

hadley:
Can you take a look at the CI failures please?

mgirlich:
Ah, it seems the names of capture groups are only properly supported for recent R versions. Fixed now 😄

R/content-type.R
```diff
@@ -0,0 +1,93 @@
+parse_content_type <- function(x) {
+  # Create regex with {rex} package
+  #
+  # ```
+  # library(rex)
+  # types <- c("application", "audio", "font", "example", "image", "message", "model", "multipart", "text", "video")
```
hadley, line 6: Add comment with source of these types?

R/resp-body.R
```diff
@@ -65,9 +65,8 @@ resp_body_string <- function(resp, encoding = NULL) {
 resp_body_json <- function(resp, check_type = TRUE, simplifyVector = FALSE, ...) {
   check_response(resp)
   check_installed("jsonlite")
-  check_content_type(resp,
+  check_resp_content_type(resp,
```
hadley, line 68: Why can the suffix arguments go away now?

mgirlich: It isn't needed as the suffix can be determined from the `types` argument.

hadley: I think that only coincidentally works for json and xml, but isn't generally true: https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml.

mgirlich: I don't see a counter example there. Maybe I misunderstood.
There is also an [extensive list of media types](https://www.iana.org/assignments/media-types/media-types.xml).

Do you think any of these cases should be treated differently?

``` r
devtools::load_all("~/GitHub/httr2/")
#> ℹ Loading httr2

# anything with suffix `+cbor` should be valid
check_content_type("application/cbor", "application/cbor")
#> NULL
check_content_type("application/ace+cbor", "application/cbor")
#> NULL
check_content_type("application/ace", "application/cbor")
#> Error:
#> ! Unexpected content type 'application/ace'
#> ℹ Expecting 'application/cbor' or 'application/<subtype>+cbor'

#> Backtrace:
#>     ▆
#>  1. └─httr2:::check_content_type("application/ace", "application/cbor")
#>  2.   └─rlang::abort(...) at httr2/R/content-type.R:68:2

# if `valid_type` has <subtype>+<suffix> the content type must have the same structure
check_content_type("application/ace+cbor", "application/ace+cbor")
#> NULL
check_content_type("application/ace", "application/ace+cbor")
#> Error:
#> ! Unexpected content type 'application/ace'
#> ℹ Expecting 'application/ace+cbor'

#> Backtrace:
#>     ▆
#>  1. └─httr2:::check_content_type("application/ace", "application/ace+cbor")
#>  2.   └─rlang::abort(...) at httr2/R/content-type.R:68:2
check_content_type("application/cbor", "application/ace+cbor")
#> Error:
#> ! Unexpected content type 'application/cbor'
#> ℹ Expecting 'application/ace+cbor'

#> Backtrace:
#>     ▆
#>  1. └─httr2:::check_content_type("application/cbor", "application/ace+cbor")
#>  2.   └─rlang::abort(...) at httr2/R/content-type.R:68:2
```

<sup>Created on 2023-01-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hadley: As far as I can see from reading the spec, there's no relationship between the subtype and the suffix, _except_ that there happens to be both subtypes and suffixes for `xml` and `json`. 

IOTW this is incorrect because there's no `+plain` suffix:

```R
check_content_type("application/cbor", "text/plain")
#> Error:
#> ! Unexpected content type 'application/cbor'
#> ℹ Expecting 'text/plain' or 'text/<subtype>+plain'
```

R/resp-body.R
```diff
@@ -103,37 +101,19 @@ resp_body_xml <- function(resp, check_type = TRUE, ...) {
 
 # Helpers -----------------------------------------------------------------
 
-check_content_type <- function(
-    resp,
-    types,
-    suffix = NULL,
-    check_type = TRUE) {
-
+check_resp_content_type <- function(resp,
```
hadley, line 104: I think we should probably export this

R/resp-body.R
hadley: I think you're pre-supposing a relationship between `type` and `suffix` that doesn't exist, and it would be better to keep `suffix` as a separate argument. In practice, I don't think this will affect how the code works, but I don't like writing a function that implies a structure that I don't believe to be true.

But I don't think I'm doing a good job of explaining exactly what I mean here, so would you mind just trusting me and switching back to an explicit `suffix` argument? I think it's a pretty simple change.

mgirlich: I'm happy to go back to `suffix`. But I'm not quite sure what the interface should like like in the end. I think there are three types of questions on the content type:

* Is the content in JSON format?
* Is the content type `application/json`?
* Is the content type `application/vnd.api+json`?

|                            | Is the content in JSON format? | Is the content type `application/json`? | Is the content type `application/vnd.api+json`? |   |
|----------------------------|--------------------------------|-----------------------------------------|-------------------------------------------------|---|
| `application/cbor`         | no                             | no                                      | no                                              |   |
| `example/json`             | yes                            | no                                      | no                                              |   |
| `application/json`         | yes                            | yes                                     | no                                              |   |
| `application/test+json`    | yes                            | yes                                     | no                                              |   |
| `application/vnd.api+json` | yes                            | yes                                     | yes                                             |   |

Do you agree that these three questions all make sense? And what do you think the interface should look like then?

jonthegeek:
It looks like the only remaining issue in this is a slight conflict in NEWS.md.

R/resp-body.R
hadley: Let take a stab at this PR and give you something to look at.

R/req-body.R
```diff
@@ -216,23 +221,28 @@ req_body_apply <- function(req) {
...
     data <- unobfuscate(data)
-    content_type <- "application/x-www-form-urlencoded"
     req <- req_body_apply_raw(req, query_build(data))
   } else {
     abort("Unsupported request body `type`", .internal = TRUE)
   }
 
-  # Must set header afterwards
+  # Respect existing Content-Type if set
+  type_idx <- match("content-type", tolower(names(req$headers)))
```
hadley, line 237: I've changed the logic a bit here — now if you manually override with the `content-type` header, we don't check it. I think this is ok, and gives the user an escape hatch if they're having problems. But generally we expect people to set the content type via the appropriate `req_body_*` function, which will verify that the type is as expected.

R/content-type.R
```diff
@@ -0,0 +1,115 @@
...
+#'   have the expected content type.
+#' @export
+#' @examples
+#' resp <- response(headers = list(`content-type` = "application/json"))
+#' check_resp_content_type(resp, "application/json")
+#' try(check_resp_content_type(resp, "application/xml"))
+#'
+#' # `types` can also specify multiple valid types
+#' check_resp_content_type(resp, c("application/xml", "application/json"))
+check_resp_content_type <- function(resp,
```
hadley, line 23: I wonder if this should be `resp_check_content_type()`; the `resp`/`req` prefix feels very strong in httr2.

R/content-type.R
```diff
@@ -0,0 +1,115 @@
+#' Check the content type of a response
+#'
+#' A different content type than expected often leads to an error in parsing
+#' the response body. This function checks that the content type of the response
+#' is as expected and fails otherwise.
+#'
+#' @param resp A response object.
+#' @param valid_types A character vector of valid content types.
+#' @param valid_suffix A string given an "structured media type" suffix.
```
hadley, line 9: This keeps the type and suffixes orthogonal in the way I think matches the spec.

hadley:
@mgirlich let me know what you think of this approach.

mgirlich:
I find it a bit confusing that the following works:

```r
check_content_type(
  "audio/abc+json",
  valid_types = "application/json",
  valid_suffix = "json"
)
```

If we want to allow this then `valid_types` should be an optional argument.

R/content-type.R
```diff
@@ -0,0 +1,115 @@
...
+  #   ),
+  #   maybe("+", capture(one_or_more(none_of(";")), name = "suffix")),
+  #   maybe(";", capture(one_or_more(any), name = "parameters")),
+  #   end
+  # )
+  # unclass(regex)
+  # ```
+  stopifnot(length(x) == 1)
+  regex <- "^(?<type>application|audio|font|example|image|message|model|multipart|text|video)/(?<subtype>(?:(?:vnd|prs|x)\\.)?(?:[^+;])+)(?:\\+(?<suffix>(?:[^;])+))?(?:;(?<parameters>(?:.)+))?$"
+  if (!grepl(regex, x, perl = TRUE)) {
```
mgirlich, line 72: Maybe this should be an error instead? And we might just have a special handling when the content type is NULL or the empty string?

hadley:
Oh yeah, we can definitely make `valid_types` optional.  I don't think it's worth validating it; but I can document it and fix that test.

PR #204, opened 2023-02-07
Add a progress bar

### LINKED ISSUE #202: Add progress bar to `sys_sleep()`
Should only display in interactive contexts

hadley:
Fixes #202

### DIFF
--- a/tests/testthat/test-utils.R
+++ b/tests/testthat/test-utils.R
@@ -36,3 +36,12 @@ test_that("respects httr verbose config", {
   withr::local_options(httr_config = config)
   expect_equal(httr2_verbosity(), 1)
 })
+
+test_that("sys_sleep has progress bar", {
+  withr::local_options(
+    cli.dynamic = FALSE,
+    cli.progress_show_after = 0,
+    cli.progress_clear = FALSE
+  )
+  expect_snapshot(sys_sleep(1))
+})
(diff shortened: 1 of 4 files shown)


tests/testthat/test-utils.R
```diff
@@ -36,3 +36,12 @@ test_that("respects httr verbose config", {
   withr::local_options(httr_config = config)
   expect_equal(httr2_verbosity(), 1)
 })
+
+test_that("sys_sleep has progress bar", {
+  withr::local_options(
+    cli.dynamic = FALSE,
```
hadley, line 42: Why isn't this causing the progress bar to update once every 3s, like https://cli.r-lib.org/articles/progress-advanced.html#non-interactive-r-sessions says?

gaborcsardi: Because the thread that provides the ticks is set up when cli is loaded, and then it is not updated: https://github.com/r-lib/cli/blob/01f932adafd58a59d596b745bf88359c5cb5a969/R/onload.R#L51-L66

hadley: Hmmmm, any thoughts on how to test this then?

gaborcsardi: Maybe create a function that formats the progress bar, and test that function, instead of "live" testing a progress bar.

gaborcsardi (commented):
I think this is fine, except that the snapshot test might be unreliable. E.g. I don't really understand the empty lines and the last last `Waiting...` line.

PR #212, opened 2023-03-12
Finish sentence about `throttle_status()`

fkohrt:
Not sure if this was the intention, though.


judith-bourque:
This would close #219.

I made a similar PR #220 based on my understanding of the function, but this might be more technically accurate.

PR #214, opened 2023-03-30
add skip_if_offline() to network dependent tests

spotrh:
Fedora requires all software to build through a build system that does not have network access for security and predictability reasons. We still want to run as many tests as possible though, so this change adds the "skip_if_offline()" check from testthat to the two testthat tests in httr2 which call request_httpbin() and thus would need network access to succeed.

I did not pass an option to skip_if_offline() to check if httpbin specifically was reachable, though, if preferred, that would be easy enough to do.

### DIFF
--- a/R/test.R
+++ b/R/test.R
@@ -12,6 +12,7 @@ request_test <- function(template = "/get", ...) {
 }
 
 request_httpbin <- function(template, ...) {
+  testthat::skip_if_offline("https://httpbin.org")
   req <- request("https://httpbin.org")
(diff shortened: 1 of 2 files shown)


hadley:
If this only affects `request_httpbin()` it might be better to put the skip in that function (since it's purely for testing). That'll ensure we can't forget to use it in future tests. (There's some discussion of that technique in https://github.com/r-lib/testthat/issues/1639)

hadley:
Superseded by #225 where we eliminate httpbin.org usage all together.

Issue #176, opened 2022-10-19
Should we provide generic oob for `req_oauth_auth_code()`?


hadley:
Like gargle is doing to work around Google's deprecation of `oob`.

Should include instructions on how to publish to your own server: https://github.com/r-lib/gargle/blob/main/inst/pseudo-oob/google-callback/index.html

Also needs tweaks to `oauth_auth_code()` to opt-out of local server. Would be useful to apply some heuristics to figure out when you need the OOB flow (which gargle likely has).

JosiahParry:
Commenting for support of oob in httr2. Though it isn't too difficult to fake it with what is already available. 

related to https://github.com/r-lib/httr/issues/617


```r
  client <- oauth_client(...)

  auth_url <- oauth_flow_auth_code_url(client, auth_url = ".../authorize", redirect_uri = "urn:ietf:wg:oauth:2.0:oob")
  
  browseURL(auth_url)
    cli::cli_alert_info("Paste authorization code:")
    token <- readline()
```

PR #248, opened 2023-07-05
Improve control over redirects during auth code flows

### LINKED ISSUE #165: Can't provide custom `redirect_uri` in `oauth_auth_code()`

atheriel:
When sheparding a user through an interactive authorisation code flow to access APIs protected by OAuth2, the `oauth_flow_auth_code()` function runs a temporary webserver on `localhost` to capture the code. This is a very common technique for OAuth client libraries.

Unfortunately, this doesn't work in environments where `localhost` is probably inaccessible to users -- which is normally the case on RStudio Server and Posit Workbench. Previously the `oauth_flow_auth_code()` function would hang indefinitely in these environments.

This commit fixes this failure mode by attempting to detect when this temporary `localhost` server will actually work and only runs one if it will.

Of course, this doesn't resolve the issue for users that can't run a `localhost` server -- so in this situation, we now allow users to manually paste the authorisation code into the console from... somewhere. In the likely event that they don't want to be redirected to `localhost` for this step, this commit also fulfills the longstanding request for control over the redirect URL.

The combination of a manual copy & paste step with a custom redirect URL is what the gargle package terms the "pseudo out-of-band" flow, and the design used here is largely based on that package.

Because of this compatibility, users could register the existing public page at https://www.tidyverse.org/google-callback/ with their OAuth provider, host their own clone, or create a variant.

Control over the redirect URL is backwards-compatible but does require making the `host_name` parameter deprecated, unfortunately.

Unit tests for new functions are included, as are various updates to the documentation.

Fixes #165 and #176.


R/oauth-flow-auth-code.R
```diff
@@ -85,14 +87,18 @@ req_oauth_auth_code <- function(req, client,
...
+#' this is not possible -- for example, when running on a hosted platform like
+#' RStudio Server -- it prompts the user to enter the code manually instead.
+#'
+#' The remaining low-level functions can be used to assemble a custom flow for
+#' APIs that are further from the spec:
 #'
 #' * `oauth_flow_auth_code_url()` generates the url where the user is sent.
-#' * `oauth_flow_auth_code_listen()` starts an webserver that listens for
-#'   the response from the resource server.
+#' * `oauth_flow_auth_code_listen()` starts a temporary local webserver that
```
jennybc, line 100: I can't comment on the line just above because it's not in the diff, but it's another example of potentially confusing phrasing about "the user".

jennybc (commented):
Exciting!

I feel like we should discuss this recommendation:

> Because of this compatibility, users could register the existing public page at https://www.tidyverse.org/google-callback/ with their OAuth provider, host their own clone, or create a variant.

I've gotten the question over in gargle land whether it's OK for other folks to use https://www.tidyverse.org/google-callback/ and, while I can't see any reason to say "no, please don't", it also feels a bit odd. How do we feel about that? 🤔

atheriel:
You're right, I don't feel like we can recommend using the `gargle` site in its current incarnation. The language on that page is very narrowly scoped to its target audience. I mentioned it mostly for demo purposes.

It's also my hope that the upcoming Workbench changes for redirects will make this pseudo-OOB workflow more uncommon -- it's really here as a last resort.

(As an aside: compatibility with `gargle` here might make it easier to refactor `gargle` to use `httr2` one day.)

R/oauth-flow-auth-code.R
```diff
@@ -85,14 +87,18 @@ req_oauth_auth_code <- function(req, client,
 #' Section 4.1. This is the most commonly used OAuth flow where the user is
 #' opens a page in their browser, approves the access, and then returns to R.
 #'
-#' `oauth_flow_auth_code()` is a high-level wrapper that should
-#' work with APIs that adhere relatively closely to the spec. The remaining
-#' low-level functions can be used to assemble a custom flow for APIs that are
-#' further from the spec:
+#' `oauth_flow_auth_code()` is a high-level wrapper that should work with APIs
+#' that adhere relatively closely to the spec. When possible, it redirects users
```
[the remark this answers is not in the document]
atheriel: What about "redirects the browser"?

R/oauth-flow-auth-code.R
```diff
@@ -296,3 +325,47 @@ oauth_flow_auth_code_pkce <- function() {
...
+  }, error = function(e) {
+    list(
+      code = code,
+      state = trimws(read_line("Enter state parameter: "))
+    )
+  })
+  if (!identical(result$state, state)) {
+    abort("Authentication failure: state does not match")
+  }
+  code <- result$code
```
hadley, line 365:
```suggestion
  result$code
```
?

R/oauth-flow-auth-code.R
```diff
@@ -296,3 +325,47 @@ oauth_flow_auth_code_pkce <- function() {
...
+  #
+  # 1) The original {gargle} style, where the user copy & pastes a
+  #    base64-encoded JSON object with both the code and state. This is used on
+  #    https://www.tidyverse.org/google-callback/; and
+  #
+  # 2) The full manual approach, where the code and state are entered
+  #    independently.
+  result <- tryCatch({
+    jsonlite::fromJSON(rawToChar(openssl::base64_decode(code)))
+  }, error = function(e) {
```
hadley, line 356:
```suggestion
  result <- tryCatch(
   jsonlite::fromJSON(rawToChar(openssl::base64_decode(code))),
   error = function(e) {
```

R/oauth-flow-auth-code.R
```diff
@@ -139,12 +149,23 @@ oauth_flow_auth_code <- function(client,
...
-                                 port = httpuv::randomPort()
+                                 port = httpuv::randomPort(),
+                                 redirect_url = "http://localhost"
 ) {
   oauth_flow_check("authorization code", client, interactive = TRUE)
-  check_installed("httpuv")
+
+  # For backwards compatibility, fall back to the original redirect URL
+  # construction.
+  if (!is.null(host_name)) {
```
hadley, line 161: Can you formally deprecate `host_name`? (https://lifecycle.r-lib.org/articles/communicate.html#deprecate-an-argument-keeping-the-existing-default)

tests/testthat/test-oauth-flow-auth-code.R
```diff
@@ -0,0 +1,38 @@
...
+  # under it.
+  withr::with_envvar(c("RSTUDIO_PROGRAM_MODE" = NA), {
+    expect_false(is_hosted_session())
+  })
+})
+
+test_that("authorisation codes can be input manually", {
+  state <- base64_url_rand(32)
+
+  # Test {gargle}-style base64-encoded JSON inputs.
```
hadley, line 15: Split into two tests?

R/oauth-flow-auth-code.R
```diff
@@ -110,12 +116,16 @@ req_oauth_auth_code <- function(req, client,
...
 #' @param token_params List containing additional parameters passed to the
 #'   `token_url`.
-#' @param host_name Host name used to generate `redirect_uri`
-#' @param host_ip IP address web server will be bound to.
-#' @param port Port to bind web server to. By default, this uses a random port.
-#'   You may need to set it to a fixed port if the API requires that the
+#' @param host_name Deprecated. Use `redirect_url` instead.
+#' @param host_ip IP address for the temporary webserver used to capture the
+#'   authorization code.
+#' @param port Port to bind the temporary webserver to. Used only when
```
hadley, line 122: Could we just extract this from the `redirect_url`? I'm actually somewhat surprised that the random port works at all, because I thought the redirect url matching was quite strict.

Overall something no longer feels quite right about the factoring of these arguments, but I'm not quite sure how to change it. 

Maybe we can drop `host_ip` since it seems unlikely that anything other than 127.0.0.1 will work?

R/oauth-flow-auth-code.R
```diff
@@ -154,9 +175,8 @@ oauth_flow_auth_code <- function(client,
   }
 
   state <- base64_url_rand(32)
-  redirect_url <- paste0("http://", host_name, ":", port, "/")
 
-  # Redirect user to authorisation url, and listen for result
+  # Redirect user to authorisation url.
```
hadley, line 179:
```suggestion
  # Redirect user to authorisation url
```

R/oauth-flow-auth-code.R
```diff
@@ -85,14 +87,18 @@ req_oauth_auth_code <- function(req, client,
...
 #'
-#' `oauth_flow_auth_code()` is a high-level wrapper that should
-#' work with APIs that adhere relatively closely to the spec. The remaining
-#' low-level functions can be used to assemble a custom flow for APIs that are
-#' further from the spec:
+#' `oauth_flow_auth_code()` is a high-level wrapper that should work with APIs
+#' that adhere relatively closely to the spec. When possible, it redirects users
+#' back to a temporary local webserver to capture the authorization code. When
+#' this is not possible -- for example, when running on a hosted platform like
+#' RStudio Server -- it prompts the user to enter the code manually instead.
```
hadley, line 94: But the package author needs to do something else to make this work, right? Maybe we should replicate the gargle approach on an httr2 web page so we can recommend that as the default?

hadley (approved):
Looks great! 

I think if we continue further down this road, we'll probably want a vignette on the pseudo-OOB approach.

R/oauth-flow-auth-code.R
atheriel: * Redirect URL matching is _generally_ strict about ports except in the very limited case of `localhost`, which some providers are lax about so that local development is easier (examples that I can find are Keycloak, Pingfederate, GitHub, possibly Azure AD).
* I'm also surprised that `host_ip` exists. I think it may have originally been to allow for `0.0.0.0` so that you could, in theory, do something like access an RStudio Server instance on `rs.acme.internal:1410` with `host_name = "rs.acme.internal", port = 1410, host_ip = "0.0.0.0"`. This would be pretty limited compared to specifying the redirect URL, though, and assumes a lot about firewalls. I've never tried it.

R/oauth-flow-auth-code.R
atheriel: We'd still need to get people to register that standard `httr2` web page with their OAuth2 providers. The reason that `gargle` works so well is that it ships with a default client that has https://www.tidyverse.org/google-callback/ preregistered.

R/oauth-flow-auth-code.R
atheriel: ... saved by invisible return of `'<-'()` there, whoops.

R/oauth-flow-auth-code.R
atheriel: Also: one big limitation of the `host_name = "rs.acme.internal", port = 1410, host_ip = "0.0.0.0"` approach is that it can't be used to specify HTTPS, and _most_ OAuth2 providers won't let you redirect to a non-HTTPS URL (other than, again, the special `localhost` case).

Overall it felt like the right approach to just add `redirect_url` as an argument rather than trying to expose enough other knobs that it would get generated correctly.

R/oauth-flow-auth-code.R
```diff
@@ -110,12 +116,16 @@ req_oauth_auth_code <- function(req, client,
...
-#'   You may need to set it to a fixed port if the API requires that the
+#' @param host_name Deprecated. Use `redirect_url` instead.
+#' @param host_ip IP address for the temporary webserver used to capture the
+#'   authorization code.
+#' @param port Port to bind the temporary webserver to. Used only when
+#'   `redirect_url` is `"http(s)://localhost"`. By default, this uses a random
+#'   port. You may need to set it to a fixed port if the API requires that the
 #'   `redirect_uri` specified in the client exactly matches the `redirect_uri`
 #'   generated by this function.
+#' @param redirect_url URL to redirect back to after authorization is complete.
```
atheriel, line 127: @hadley There is an existing `redirect_uri` argument on this help page (for `oauth_flow_auth_code_url()`) -- should I aim for consistency and use that naming convention here as well?

R/oauth-flow-auth-code.R
hadley: I agree that `redirect_uri` is the right thing to expose; I was wondering if we should move to eliminate some of the old knobs.

R/oauth-flow-auth-code.R
hadley: Yeah, we should make them consistent.

R/oauth-flow-auth-code.R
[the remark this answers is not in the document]
atheriel: Updated the wording.

R/oauth-flow-auth-code.R
atheriel: Updated the wording.

R/oauth-flow-auth-code.R
atheriel: Done.

R/oauth-flow-auth-code.R
atheriel: Done.

tests/testthat/test-oauth-flow-auth-code.R
atheriel: Done.

NEWS.md
```diff
@@ -13,6 +13,13 @@
 * `oauth_flow_refresh()` now only warns if the `refresh_token` changes, making
   it a little easier to use in manual workflows (#186).
 
+* `oauth_flow_auth_code()` now attempts to detect when a temporary webserver can
+  be run on `localhost`. If not, it allows users to enter the authorisation code
```
hadley, line 17: I think it would be good to be explicit about Collab/Workbench/Cloud since otherwise this sounds like it detects if you cant start a web server.

hadley (approved):

R/oauth-flow-auth-code.R
```diff
@@ -154,26 +184,34 @@ oauth_flow_auth_code <- function(client,
...
+    redirect_uri = redirect_uri,
     scope = scope,
     state = state,
     auth_params = auth_params
   )
   utils::browseURL(user_url)
-  result <- oauth_flow_auth_code_listen(host_ip, port)
-  code <- oauth_flow_auth_code_parse(result, state)
+
+  if (!is_hosted_session() && is_installed("httpuv")) {
```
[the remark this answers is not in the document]
atheriel: I agree that it's a bit odd for behaviour to vary, I was aping `gargle` here: https://github.com/search?q=repo%3Ar-lib%2Fgargle+httpuv+language%3AR&type=code&l=R

R/oauth-flow-auth-code.R
```diff
@@ -85,14 +87,20 @@ req_oauth_auth_code <- function(req, client,
...
 #'
-#' `oauth_flow_auth_code()` is a high-level wrapper that should
-#' work with APIs that adhere relatively closely to the spec. The remaining
-#' low-level functions can be used to assemble a custom flow for APIs that are
-#' further from the spec:
+#' `oauth_flow_auth_code()` is a high-level wrapper that should work with APIs
+#' that adhere relatively closely to the spec. When possible, it redirects the
+#' browser back to a temporary local webserver to capture the authorization
+#' code. When this is not possible -- for example, when running on a hosted
+#' platform like RStudio Server -- it prompts the user to enter the code
```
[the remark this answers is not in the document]
atheriel: Sad.

R/oauth-flow-auth-code.R
```diff
@@ -133,18 +146,35 @@ req_oauth_auth_code <- function(req, client,
...
                                  scope = NULL,
                                  pkce = TRUE,
                                  auth_params = list(),
                                  token_params = list(),
-                                 host_name = "localhost",
+                                 host_name = deprecated(),
                                  host_ip = "127.0.0.1",
-                                 port = httpuv::randomPort()
+                                 port = httpuv::randomPort(),
+                                 redirect_uri = "http://localhost"
```
hadley, line 159: I wonder if we should deprecate `port` too, replacing this default with something that automatically adds the random port.

hadley: Oh that's what I said above, previously 😆. Let's leave this as is for this PR and consider a refactoring in a future PR: #291.

R/oauth-flow-auth-code.R
```diff
@@ -56,9 +56,11 @@ req_oauth_auth_code <- function(req, client,
                                 pkce = TRUE,
                                 auth_params = list(),
                                 token_params = list(),
+                                type = c("desktop", "web"),
```
atheriel, line 59: For consistency, maybe we should use type `"hosted"` here instead of type `"web"`.

hadley: It occurred to me that maybe we could do this automatically based on whether or not the `redirect_url` was to localhost?

hadley: Let me merge this and then present a concrete proposal.

PR #263, opened 2023-08-03
Add argument `.redact` to `req_headers()`

### LINKED ISSUE #247: Support user-defined redaction in `req_headers()`
The Google Cloud APIs use the [`x-goog-api-key` header](https://cloud.google.com/docs/authentication/api-keys#using-with-rest) to pass API keys. Currently, only the `Authorization` header is automatically redacted by `req_headers()`, as implemented in the internal function [headers_redact()](https://github.com/r-lib/httr2/blob/66c07ded035ae33ce2ed27f2c79a817286bcb1e8/R/headers.R#L44). `x-goog-api-key` is therefore not redacted:

``` r
library(httr2)

req_google <-
  request("https://www.googleapis.com") %>%
  req_url_path_append("youtube/v3") %>%
  req_headers("x-goog-api-key" = "ABC123XYZ")

req_google
#> <httr2_request>
#> GET https://www.googleapis.com/youtube/v3
#> Headers:
#> • x-goog-api-key: 'ABC123XYZ'
#> Body: empty
```
<sup>Created on 2023-06-29 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I recently added support for YouTube Data API key authentication in https://github.com/soodoku/tuber/pull/126, and it would be great to be able to redact the `x-goog-api-key` header used [here](https://github.com/soodoku/tuber/pull/126/files#diff-e746f18382d462bf758a3eb6f41f4c3e7dff6fafe019f17a5fab3f0d0dc75b29R180) for use in continuous integration. More broadly, a `.redact = TRUE` argument to `req_headers()` would permit users to redact any header. The desired behavior would look like the code block below:

``` r
library(httr2)

req_google <-
  request("https://www.googleapis.com") %>%
  req_url_path_append("youtube/v3") %>%
  req_headers("x-goog-api-key" = "ABC123XYZ", .redact = TRUE)

req_google
#> <httr2_request>
#> GET https://www.googleapis.com/youtube/v3
#> Headers:
#> • x-goog-api-key: '<REDACTED>'
#> Body: empty
```

mgirlich:
Closes #247.
The reason to take a named list instead of a named logical is to be consistent with e.g. the `.simplify` argument of `tidyr::hoist()`.


R/req-headers.R
```diff
@@ -9,6 +9,10 @@
 #'   * Use `NULL` to reset a value to httr's default
 #'   * Use `""` to remove a header
 #'   * Use a character vector to repeat a header.
+#' @param .redact If `TRUE`, all the added headers are redacted. Can also be a
```
hadley, line 12: What about making this a character vector? It makes it slightly harder to use in the `.redact = TRUE` case, but is easier if you only want to redact a single header, and I think a storing character vector will make for a much simpler implementation.

hadley: I don't think supporting `TRUE` is needed IMO. I think it will most be used by package authors, and it's nice to be explicit.

R/headers.R
```diff
@@ -37,11 +37,14 @@ print.httr2_headers <- function(x, ..., redact = TRUE) {
   invisible(x)
 }
 
-headers_redact <- function(x, redact = TRUE) {
+headers_redact <- function(x, redact = TRUE, to_redact = NULL) {
   if (!redact) {
     x
   } else {
-    list_redact(x, "Authorization", case_sensitive = FALSE)
+    to_redact <- union(attr(x, "redact"), c("Authorization", to_redact = to_redact))
```
mgirlich, line 44: `Authorization` is added manually for the tests to keep working. Do you want to keep it that way to help users avoid accidentally showing credentials? Or should we remove it as it is added anyway in `req_auth_basic()` and `req_auth_bearer_token()`?

R/req-headers.R
```diff
@@ -47,9 +49,31 @@
...
+  redact_out <- attr(.req$headers, "redact") %||% .redact %||% character()
+  redact_out <- union(redact_out, .redact)
+  .req$headers <- modify_list(.req$headers, !!!headers)
+
+  attr(.req$headers, "redact") <- redact_out
+
   .req
 }
+
+check_redact <- function(x, arg = caller_arg(x), call = caller_env()) {
```
mgirlich, line 72: It might make sense to add the rlang standalone file for type checks in a separate PR.

R/req-options.R
```diff
@@ -197,13 +198,13 @@ verbose_message <- function(prefix, x) {
   cli::cat_line(prefix, lines)
 }
 
-verbose_header <- function(prefix, x, redact = TRUE) {
+verbose_header <- function(prefix, x, redact = TRUE, to_redact = NULL) {
```
mgirlich, line 201: `to_redact` is needed as `line` is parsed into a header and therefore doesn't have the `redact` attribute.

R/req-perform.R
```diff
@@ -212,8 +212,9 @@ req_dry_run <- function(req, quiet = FALSE, redact_headers = TRUE) {
   check_installed("httpuv")
 
   if (!quiet) {
+    to_redact <- attr(req$headers, "redact")
     debug <- function(type, msg) {
-      if (type == 2L) verbose_header("", msg, redact = redact_headers)
+      if (type == 2L) verbose_header("", msg, redact = redact_headers, to_redact = to_redact)
```
mgirlich, line 217: I'm not sure that `req_dry_run()` should redact headers but overall it feels more consistent.

R/headers.R
hadley: I think as long as it's added by the calls that actually use the `Authorization` header it should be ok.

hadley (approved):

PR #265, opened 2023-08-08
Use rlang standalone type check

mgirlich:


R/oauth-token.R
```diff
@@ -15,7 +15,7 @@
 #' @export
 #' @examples
 #' oauth_token("abcdef")
-#' oauth_token("abcdef", expires_in = Sys.time() + 3600)
```
hadley, line 18: Nice catch

PR #272, opened 2023-08-09
Give more info in progress bars

### LINKED ISSUE #206: `req_perform()` should give details about why it's waiting
Is it throttling? Is it a back-off or information from header? How many times has it tried and how many retries are left?

hadley:
Fixes #206


R/utils.R
```diff
@@ -40,8 +40,8 @@ modify_list <- function(.x, ..., error_call = caller_env()) {
 }
 
 
-sys_sleep <- function(seconds, fps = 10, error_call = caller_env()) {
-  check_number_decimal(seconds, call = error_call)
+sys_sleep <- function(seconds, task, fps = 10) {
```
hadley, line 43: @mgirlich I reverted the use of `error_call` here because I don't think it's useful as `seconds` is not directly passed to `sys_sleep()`. This is more of an internal fallback to ensure that `seconds` is the right type; validation of user inputs should happen earlier.

PR #279, opened 2023-08-15
Draft: Pagination support

mgirlich:
Pagination support would be great to have. This PR provides basic support for pagination. It adds a general `req_paginate()` and three helpers for common patterns:

* `req_paginate_next_url()`: the response contains a next url (usually in the body)
* `req_paginate_offset()`: simple pagination with `limit` and `offset` (typically in the query)
* `req_paginate_next_token()`: pagination via a token/cursor

I could test the first two helpers via the public Pokemon API, the last one only via an API that needs authorisation.

```r
# next url ----------------------------------------------------------------

library(tibblify)
responses <- request("https://pokeapi.co/api/v2/pokemon") %>% 
  req_paginate_next_url(
    next_field = "next",
    results_field = "results",
    limit = in_query("limit", 150L),
    total_field = "count",
    max_pages = 2
  )

spec <- tspec_df(
  tib_chr("name"),
  tib_chr("url"),
)

pokemon_df <- purrr::map(
  responses,
  \(resp_body) {
    tibblify(resp_body, spec)
  }
) %>% 
  purrr::list_rbind()


# offset ------------------------------------------------------------------

responses <- request("https://pokeapi.co/api/v2/pokemon") %>% 
  req_paginate_offset(
    offset = in_query("offset"),
    results_field = "results",
    limit = in_query("limit", 150L),
    total_field = "count",
    max_pages = 20
  )

pokemon_df <- purrr::map(
  responses,
  \(resp_body) {
    tibblify(resp_body, spec)
  }
) %>% 
  purrr::list_rbind()
```

* this doesn't return the full response but only the body. I guess in most cases the body is more relevant than the full response. This removes some clutter code from the user. Also, some pagination pattern require parsing the body anyway.
* this currently focuses on a JSON body for simplicity. But this should already cover the most typical case for pagination.
* this doesn't support returning intermediate results e.g. in case of an error (for interactive purposes it would also be nice to have intermediate results after an user interrupt).


mgirlich:
@hadley Would be great to get feedback on this so that we can add some basic pagination support to httr2 soon 😄

hadley:
Looking at this very quickly, I wonder if we should start a bit lower level so if needed you get greater control over the process. Maybe something like this?

```R
library(tibblify)
req <- request("https://pokeapi.co/api/v2/pokemon") %>% 
  req_paginate(
    next_field = "next",
    results_field = "results",
    limit = in_query("limit", 150L),
    total_field = "count"
  )

repeat({
  resp <- req_perform(req)
  req <- resp_next_request(resp)
  if (is.null(req)) break
})
```

Once we have that in place, it would be trivial to build these one-size-fits-most helpers on top.

I'm not sure what `req_paginate()` should look like, but I suspect it should probably start with a callback that takes a request and a response, so you can modify the request using bits from the response.

mgirlich:
Good point with the lower level version. The api now consists of:

* `req_paginate()` which adds pagination policies to the request
* `req_paginate_next_url()`, `req_paginate_offset()`, `req_paginate_next_token()` which are helpers for special cases
* `paginate_next_request()` to create the request to the next page
* `paginate_n_pages()` to calculate the total number of pages
* `paginate_perform()` to perform the pagination requests and return a list of responses

There are also three functions needed that specify where in the request a parameter is set:

* `in_query()`: add a parameter as query
* `in_body()`: add a parameter in the body
* `in_header()`: not sure if this is actually needed but I added it for the sake of completeness

They are needed in two ways:

```r
request("https://pokeapi.co/api/v2/pokemon") %>% 
  req_paginate_offset(
    offset = in_query("offset"),
    page_size = in_query("limit", 150L),
    total = "count"
  )
```

The page size often isn't needed and could also be set before by the user. The advantage of having that in `req_paginate()` is that the total number of available pages can be calculated (in case the response contains the total number of elements).
The `offset` needs to be manipulated on every page.

R/paginate.R
```diff
@@ -0,0 +1,331 @@
...
+                     error_call = caller_env()) {
+  check_string(name, call = error_call)
+  out <- list(value = value, name = name)
+  class(out) <- c("httr2_query_param", "httr2_param")
+  out
+}
+
+#' @rdname in_query
+#' @export
+in_header <- function(name,
```
mgirlich, line 258: I didn't encounter an example where pagination information is passed via the header but this probably exists (and also makes sense to me)

R/paginate.R
```diff
@@ -0,0 +1,331 @@
...
+#' request("https://pokeapi.co/api/v2/pokemon") %>%
+#'   req_paginate_next_url(
+#'     "next",
+#'     page_size = in_query("limit", 150L),
+#'     total = "count"
+#'   )
+req_paginate <- function(req,
+                         next_request,
+                         page_size = NULL,
+                         total = NULL) {
```
mgirlich, line 36: Probably the `total` interface isn't flexible enough. The response might contain a) the number of elements (what I assume in this implementation) or b) the number of pages.

R/paginate.R
```diff
@@ -0,0 +1,331 @@
...
+#' @export
+req_paginate_next_url <- function(req,
+                                  next_url,
+                                  ...,
+                                  page_size = NULL,
+                                  total = NULL) {
+  check_character(next_url)
+
+  next_request <- function(req, resp) {
+    body_parsed <- resp_body_json(resp)
```
mgirlich, line 161: It would be nice to cache the parsed body. Otherwise the body is usually parsed twice.

R/paginate.R
```diff
@@ -0,0 +1,331 @@
...
+#' @rdname req_paginate
+#' @export
+req_paginate_offset <- function(req,
+                                offset,
+                                page_size,
+                                total = NULL) {
+  check_param(offset, allow_null_value = TRUE)
+  check_param(page_size)
+
+  cur_offset <- 0L
```
mgirlich, line 188: This doesn't work if `paginate_perform()` is executed twice on the same request. This indicates that we might need to add an `init` callback to `api_paginate()`.

R/paginate.R
```diff
@@ -0,0 +1,336 @@
...
+#' req_pokemon <- request("https://pokeapi.co/api/v2/pokemon") %>%
+#'   req_paginate_next_url(
+#'     "next",
+#'     page_size = in_query("limit", 150L),
+#'     total = "count"
+#'   )
+#'
+#' responses <- paginate_perform(req_pokemon)
+paginate_perform <- function(req,
+                             max_pages = 20L,
```
mgirlich, line 73: It probably makes sense to add a callback `data` that is applied on the response and returns the data to store.

R/paginate.R
hadley: https://docs.github.com/en/rest/guides/using-pagination-in-the-rest-api?apiVersion=2022-11-28#using-link-headers maybe?

R/paginate.R
```diff
@@ -0,0 +1,336 @@
...
+#' @param name Name of the query resp. header parameter.
+#' @param path Path of the parameter in the body.
+#' @param value The value of the parameter.
+#' @param error_call
+#'
+#' @return A parameter object.
+#' @export
+#'
+#' @examples
+#' in_query("start", 20)
```
hadley, line 249: I wonder if we need these helpers. It's not too awful to supply an anonymous function if there's an existing response helper:

```R
page_size = \(resp) resp_url_query(resp, "limit", 150L),
next_page = \(resp) resp_link_url(resp, "next")
```

Then we'd just need something extra in `resp_body_json()` that let you drill down to a specific component.

hadley: (That makes me wonder if `resp_link_url()` should actually be `resp_header_link()`)

jonthegeek:
This will be sooooooooo useful!

Is there anything I can do to help? If it's ready to kick the tires, I can try it out on a few APIs.

mgirlich:
> This will be sooooooooo useful!
> 
> Is there anything I can do to help? If it's ready to kick the tires, I can try it out on a few APIs.

I would be happy if you try it out and provide feedback.

R/paginate.R
```diff
@@ -0,0 +1,229 @@
...
+#'
+#' @param page_size An integer.
+#' @param total A function that extracts the total number of elements from the [response].
+#'
+#' @return A function that can be passed to the `n_pages` argument of [req_paginate()].
+#' @export
+#'
+#' @examples
+#' calculate_n_pages(page_size = 100, total = 250)
+calculate_n_pages <- function(page_size, total) {
```
mgirlich, line 221: Before this took a response object and `total` was a function that takes the response as an argument.
In this form it is hardly worth exporting. On the other hand when doing pagination one usually gets either the total number of elements (and needs to know the page size) or the total number of pages.

hadley:
@mgirlich looking at the interface now, I'd say in a future PR we should add an argument for parsing the body so the interface could look something like this:

```R
req_next_url <- request("https://pokeapi.co/api/v2/pokemon") %>% 
  req_url_query(limit = 150) %>% 
  req_paginate_next_url(
    body = \(resp) resp_body_json(resp),
    next_url = \(resp, body) body[["next"]],
    n_pages = \(resp, body) ceiling(body$count / page_size)
  )
```

I'd suggest we don't expect `calculate_n_pages()` for now, and similarly worry about the the repeated `page_size` later.

R/paginate.R
```diff
@@ -0,0 +1,229 @@
+#' Pagination
+#'
```
hadley, line 2: Needs a little description here.

R/paginate.R
```diff
@@ -0,0 +1,229 @@
...
+#' @inheritParams req_perform
+#' @param next_request A callback function that takes a two arguments (the
```
hadley, line 4: Can you move this next to the specific function that it applies to?

R/paginate.R
```diff
@@ -0,0 +1,229 @@
...
+#'   * `NULL` if there is no next page.
+#' @param page_size A parameter object that specifies how the page size is added
+#'   to the request.
+#' @param next_url A function that extracts the next url from the [response].
+#' @param offset A function that applies that applies the new offset to the
+#'   request. It takes two arguments: a [request] and an integer offset.
+#' @param set_token A function that applies that applies the new token to the
+#'   request. It takes two arguments: a [request] and the new token.
+#' @param next_token A function that extracts the next token from the [response].
+#' @param n_pages A function that extracts the next token from the [response].
```
hadley, line 17: I think this is out of date?

R/paginate.R
```diff
@@ -0,0 +1,229 @@
...
+  req_policies(
+    req,
+    paginate = list(
+      next_request = next_request,
+      n_pages = n_pages
+    )
+  )
+}
+
+#' Perform a paginated request
```
hadley, line 50: Should we document this with `req_paginate()`?

R/paginate.R
hadley: Agreed; but lets do that in the next PR.

R/paginate.R
```diff
@@ -0,0 +1,229 @@
...
+#'     n_pages = function(resp) {
+#'       calculate_n_pages(
+#'         page_size = 150,
+#'         total = resp_body_json(resp)$count
+#'       )
+#'     }
+#'   )
+#'
+#' responses <- paginate_perform(req_pokemon)
+paginate_perform <- function(req,
```
hadley, line 74: Maybe `paginate_req_perform()` to parallel `multi_req_perform()`?

R/paginate.R
```diff
@@ -0,0 +1,229 @@
...
+paginate_next_request <- function(resp, req) {
+  check_response(resp)
+  check_request(req)
+
+  if (!req_policy_exists(req, "paginate")) {
+    cli::cli_abort(c(
+      "{.arg req} doesn't have a pagination policy",
+      i = "You can add pagination via `req_paginate()`."
+    ))
+  }
```
hadley, line 129: Maybe move this check to `paginate_perform`?

R/paginate.R
```diff
@@ -0,0 +1,229 @@
...
+
+  cur_offset <- 0L
+  env <- current_env()
+  next_request <- function(req, resp) {
+    cur_offset <- get("cur_offset", envir = env)
+    new_offset <- cur_offset + page_size
+    assign("cur_offset", new_offset, envir = env)
+
+    offset(req, new_offset)
+  }
```
hadley, line 176:
```suggestion
  next_request <- function(req, resp) {
    cur_offset <<- cur_offset + page_size 
    offset(req, cur_offset)
  }
```
?

R/paginate.R
hadley: I'd say not to export this for now.

R/req-body.R
```diff
@@ -216,7 +216,8 @@ req_body_apply <- function(req) {
   } else if (type == "raw") {
     req <- req_body_apply_raw(req, data)
   } else if (type == "json") {
-    content_type <- "application/json"
+    # FIXME temporary workaround just for testing purposes. Remove before merging!
```
hadley, line 219: Need to fix now?

hadley (approved):

jonthegeek:
I got it to work with Slack 🎉🎉🎉

Some comments: 

- It took me a while to find `paginate_perform()`. If we need a separate `peform()`, `req_perform()` should tell me that I have unused pagination info, ideally, and point me to the proper performer.
- The `n_pages` help is the `next_token` help. I thought it might be my issue at first and wasn't sure what was supposed to go there.
- Multipage response helpers would be helpful, but I'm still wrapping my head around whether they make sense (since the way to combine results probably depends a lot on the particular API).

Overall this worked great, it just could use some documentation tweaks. I'm very happy to see this in action!

R/paginate.R
mgirlich: I thought it makes sense to also export `paginate_next_request()` so I left the check there but also added one to `paginate_req_perform()`.

mgirlich:
> * It took me a while to find `paginate_perform()`. If we need a separate `peform()`, `req_perform()` should tell me that I have unused pagination info, ideally, and point me to the proper performer.

To inform in `req_perform()` sounds nice but this would also imply an extra argument to `req_perform()` to switch off the information in case you actually want to use `req_perform()`. For now I simply pointed to `paginate_req_perform()` more clearly in the documentation

> * Multipage response helpers would be helpful, but I'm still wrapping my head around whether they make sense (since the way to combine results probably depends a lot on the particular API).

Let's tackle that in a separate PR. I guess at least adding an extra callback for processing the response would make sense.

PR #281, opened 2023-08-17
Implement resp_url() helpers

### LINKED ISSUE #57: Extract query parameters
I spend a few minutes to try out `httr2`. Great work with the new oauth system. I didn't really managed to use oauth with `httr` but it only took me two minutes to make it work in `httr2` 😄 

I want to extract the url query paramters of a request and wondered whether there is an easier way than the following

``` r
library(httr2)
req <- request("https://example.com/path/to/page?name=ferret&color=purple")

req_queries <- url_parse(req$url)$query
req_queries$name
#> [1] "ferret"
```

<sup>Created on 2021-07-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

hadley:
Fixes #57

Not sure if the query interface should look more like `resp_header()`/`resp_headers()`, i.e. is it useful to provide a helper to extract a single query parameter and provide a default if it's not found?


tests/testthat/test-resp-url.R
```diff
@@ -0,0 +1,7 @@
+test_that("can extract url components from a response", {
+  resp <- req_perform(request_test("/get?a=1&b=2"))
+
+  expect_equal(resp_url(resp), paste0(example_url(), "get?a=1&b=2"))
+  expect_equal(resp_url_path(resp), "/get")
+  expect_equal(resp_url_query(resp), list(a = "1", b = "2"))
```
mgirlich, line 6: Probably not really worth the complexity but this examples makes me wonder whether something like a `ptypes` argument would make sense. It would be nice to easily get `list(a = 1, b = 2)`.

mgirlich (approved):

mgirlich:
I think it would be nice to have `resp_url_queries()` and `resp_url_query()`

PR #282, opened 2023-08-21
Implement req_cookie_file()

### LINKED ISSUE #223: Share cookies across requests
I believe `httr2` uses a new handle for every request. It would be good if there was an option to retain the handle, as I have a use case where I need to reuse the same session cookies.

hadley:
Fixes #223


hadley:
@mgirlich what do you think of the name?

jonthegeek (commented):
This looks super useful! Some thoughts on the name:

- `req_cookie_file()` makes it feel like the thing I care about is the file, moreso than the reuse of the cookies.
- Would it make sense to align it with the `cache_disk` and `cache_key` options from the `_auth_` functions? It's not QUITE the same thing but feels similar. 

I guess the difference here is that you could conceivably supply a file from some other source, whereas caching is more internal and automatic.

So after talking it out... I can't think of anything that makes more sense to me than `req_cookie_file()`.

R/req-cookies.R
```diff
@@ -0,0 +1,38 @@
+#' Preserve cookies across requests
+#'
+#' By default, httr2 uses a clean slate for every request meaning that cookies
+#' are not automatically preserved across requests. To preserve cookies, you
+#' must set a cookie file which will be read before and updated after each
+#' request.
+#'
+#' @inheritParams req_perform
+#' @param path A path to file where cookies will be saved.
```
mgirlich, line 9:
```suggestion
#' @param path A path to a file where cookies will be read from before and updated after the request.
```

R/req-cookies.R
```diff
@@ -0,0 +1,38 @@
...
+#'   req_perform() %>%
+#'   resp_body_json()
+#'
+#' httpbin %>%
+#'   req_template("/cookies/set/:name/:value", name = "oatmeal", value = "raisin") %>%
+#'   req_perform() %>%
+#'   resp_body_json()
+#'
+#' # The cookie path has a straightforward format
+#' readLines(path)
```
mgirlich, line 28: maybe use `readChar()` + `cat()` for nicer printing?

```suggestion
cat(readChar(path, nchars = 1e4))
```

mgirlich (approved):

mgirlich:
Basically, `req_cookie_file()` is fine but maybe something like `req_cookie_preserve()` would be easier to remember.

hadley:
Oh I like `req_cookie_preserve()`.

hadley:
@jonthegeek the implementation here is pretty limited by what curl provides; it has to be written out to disk.

PR #285, opened 2023-08-26
Add option to suppress progress bar

### LINKED ISSUE #251: Way to turn off cli progress bar during wait
When I use req_retry or req_throttle httr2 puts up a status bar for waits longer than 1 sec. This can mess up my other output and unless I set terminal settings explicitly makes some output unreadable. 

For instance, the image below is from a throttled output along with one req_retry in the RStudio background tabs pane interspersed with my output.

Would like an option to just turn this off.

httr_output

hadley:
Fixes #251

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # httr2 (development version)
 
+* `req_performs()` progress bar can be suppressed by setting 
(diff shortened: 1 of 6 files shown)


gaborcsardi (approved):
Looks great!

PR #286, opened 2023-08-26
Add OAuth vignette & cache path customisation

### LINKED ISSUE #234: Improve docs for token caching
How does it work, where do the token live, how can you customise, etc.

hadley:
Fixes #234


R/oauth.R
```diff
@@ -116,3 +116,20 @@ cache_disk_prune <- function(days = 30, path = rappdirs::user_cache_dir("httr2")
...
+
+#' httr2 cache location
+#'
+#' When opted-in to, httr2 caches OAuth tokens in this directory. By default,
+#' it uses a OS-standard cache directory, but, if needed, you can override the
+#' location by setting the `HTTR2_CACHE` env var.
+#'
+#' @export
+#' @keywords internal
+cache_path <- function() {
```
mgirlich, line 128: I think `cache_path()` is too generic and other packages or custom script already use that name. So maybe rather `httr2_cache_path()`?

hadley: I agree that `cache_path()` seems a bit too short, but no other function has an `httr2_` prefix. Maybe `oauth_cache_path()` is a better name.

PR #289, opened 2023-08-28
Add secret_encrypt_file() and secret_decrypt_file()

### LINKED ISSUE #237: Helpers for encrypting/decrypting files
Since sometimes you need to pass a path to some other function:

```R
secret_read_file <- function(path, envir = parent.frame()) {
  val <- readChar(path, file.info(path)$size)  
  enc <- secret_encrypt(val, key = "GARGLE_KEY")
  
  path <- withr::with_tempfile(envir = envir)
  writeChar(enc, path)
  Sys.chmod(path, 400)
  path
}

secret_write_file <- function(path_in, path_out, envir = parent.frame()) {
  val <- readChar(path_in, file.info(path_in)$size)  
  enc <- secret_decrypt(val, key = "GARGLE_KEY")
  
  writeChar(enc, path_out)
  invisible(path_out)
}
```

hadley:
Fixes #237


NEWS.md
```diff
@@ -1,5 +1,8 @@
 # httr2 (development version)
 
+* New `secret_decrypt_file()` and `secret_encrypt_file()` for decrypting files
```
jennybc, line 3: Nitpick: seems like `encrypt` should come first.

R/secret.R
```diff
@@ -24,11 +26,13 @@
 #'     as an env var (e.g. `{MYPACKAGE}_KEY`) by adding a line to your
 #'     `.Renviron`.
 #'
-#' 2.  Encrypt strings with `secret_encrypt()` and other data with
-#'     `secret_write_rds()`, setting `key = "{MYPACKAGE}_KEY"`.
+#' 2.  Encrypt strings with `secret_encrypt()`, files with
+#'     `secret_encrtype_file()`,  and other data with `secret_write_rds()`,
```
jennybc, line 30:
```suggestion
#'     `secret_encrypt_file()`, and other data with `secret_write_rds()`,
```

R/secret.R
```diff
@@ -24,11 +26,13 @@
...
+#'     setting `key = "{MYPACKAGE}_KEY"`.
 #'
-#' 3.  In your tests, decrypt the data with `secret_decrypt()` or
-#'     `secret_read_rds()` to match how you encrypt it.
+#' 3.  In your tests, decrypt the data with `secret_decrypt()`,
+#'     `secret_decrypt_file()`, or `secret_read_rds()` to match how you encrypt
+#'     it.
 #'
 #' 4.  If you push this code to your CI server, it will already "work" because
 #'     all functions automatically skip tests when your `{MYPACKAGE}_KEY}`
```
jennybc, line 38: Not in this diff but there's one too many closing `}` in `` `{MYPACKAGE}_KEY}` ``

R/secret.R
```diff
@@ -113,6 +117,32 @@ secret_read_rds <- function(path, key) {
   secret_unserialize(x, key)
 }
 
+#' @export
+#' @param envir The unencrypted file will be automatically deleted when
```
jennybc, line 121: "decrypted"?

R/secret.R
```diff
@@ -113,6 +117,32 @@ secret_read_rds <- function(path, key) {
...
+#'   this environment exits. You should only need set this argument if you
```
jennybc, line 122:
```suggestion
#'   this environment exits. You should only need to set this argument if you
```

R/secret.R
```diff
@@ -113,6 +117,32 @@ secret_read_rds <- function(path, key) {
...
+#'   want to pass the unencrypted file to another function.
+#' @rdname secrets
+secret_decrypt_file <- function(path, key, envir = parent.frame()) {
+  val <- readChar(path, file.info(path)$size)
```
jennybc, line 126: Is there a particular reason for focusing on text files? I think httr2's focus on encrypting/decrypting strings and text is part of how I ended up writing `gargle::secret_(write|read)_rds()` and `gmailr::gm_token_(write|read)`.

jennybc (approved):

R/secret.R
hadley: This is just because `secret_encrypt()` returns a base64 encoded string. This isn't really necessary for files, but it seems simple enough to keep it. It shouldn't affect your ability to encrypt and decrypt binary files. (And there's already `secret_write_rds()` and `secret_read_rds()` if you want rds files).

hadley: Oh but I still need to read a binary file when encrypting. So probably better to be consistent.

hadley: And that made me realise that there's a bunch of duplicate code between string and rds encryption, so overall a nice refactoring. Thanks for spotting the problem!

PR #300, opened 2023-09-01
Add `parse_resp` to `req_paginate()`

### LINKED ISSUE #297: Add argument `body` to `req_paginate()`
@mgirlich looking at the interface now, I'd say in a future PR we should add an argument for parsing the body so the interface could look something like this:

```R
req_next_url <- request("https://pokeapi.co/api/v2/pokemon") %>% 
  req_url_query(limit = 150) %>% 
  req_paginate_next_url(
    body = \(resp) resp_body_json(resp),
    next_url = \(resp, body) body[["next"]],
    n_pages = \(resp, body) ceiling(body$count / page_size)
  )
```

I'd suggest we don't expect `calculate_n_pages()` for now, and similarly worry about the the repeated `page_size` later.

_Originally posted by @hadley in https://github.com/r-lib/httr2/issues/279#issuecomment-1701706323_

mgirlich:
Closes #297.


R/paginate.R
```diff
@@ -142,15 +157,43 @@ paginate_next_request <- function(resp, req) {
...
 #' @rdname req_paginate
 #' @export
 req_paginate_next_url <- function(req,
                                   next_url,
+                                  body = NULL,
                                   n_pages = NULL) {
   check_function2(next_url, args = "resp")
 
+  next_url_user <- next_url
+  next_url <- paginate_register_body(next_url_user)
```
mgirlich, line 195: I'm not quite sure the approach in `paginate_register_body()` makes sense as I have to modify the `next_url` function instead of only modifying the `next_request` function in `req_paginate()`

mgirlich:
For now this is only something to discuss. I'm not quite sure this is the right approach:

* how to make it transparent that the user can simply use `body` in the functions?
* is there a better implementation so that it is sufficient to modify `next_request` instead of also e.g. `next_url` in `req_paginate_next_url()`?
* maybe the variable we register should be called `.body`?

R/paginate.R
```diff
@@ -40,15 +40,26 @@
 #'   )
 req_paginate <- function(req,
                          next_request,
+                         body = NULL,
```
hadley, line 43: I was thinking this would default to `\(resp) NULL` so that `nest_request` and `n_pages` could require a `body` argument. I think that makes the rest of the implementation quite a bit simpler.

R/paginate.R
```diff
@@ -13,13 +13,18 @@
...
+#'   page or `NULL` if there is no next page. It takes a three arguments:
 #'
-#'   * a new [request] to request the next page or
-#'   * `NULL` if there is no next page.
-#' @param n_pages A function that extracts the total number of pages from
-#'   the [response].
+#'   1. `req`: the original request
+#'   2. `resp`: the response of the current request
+#'   3. `body`: the result of the argument `body` below.
+#' @param body A function with one argument `resp` that parses the body of the
```
mgirlich, line 22: This gets quite confusing. Maybe this should be renamed to `parse_body` or something. Any ideas for a good name?

hadley: Maybe the argument to `req_paginate()` should be `parse_resp`? And then the argument to `next_request` would be `parsed`?

R/paginate.R
```diff
@@ -106,15 +120,15 @@ paginate_req_perform <- function(req,
...
     if (is.null(req)) {
       page <- page - 1L
       break
     }
 
     resp <- req_perform(req)
+    body <- paginate_parse_body(resp, req)
 
-    body_parsed <- resp_body_json(resp)
     out[[page]] <- resp
```
hadley, line 132: I wonder if we should use the parsed object here too?

mgirlich: Sounds very reasonable to me. Also after renaming to `parse_resp` this sounds a) general enough and b) clear enough to kind of expect this.

R/paginate.R
```diff
@@ -13,13 +13,20 @@
...
+#'   page or `NULL` if there is no next page. It takes a three arguments:
 #'
-#'   * a new [request] to request the next page or
-#'   * `NULL` if there is no next page.
-#' @param n_pages A function that extracts the total number of pages from
-#'   the [response].
+#'   1. `req`: the original request.
+#'   2. `resp`: the response of the current request.
+#'   3. `parsed`: the result of the argument `parse_resp`.
+#' @param parse_resp A function with one argument `resp` that parses the
```
hadley, line 22: What happens if this errors?

R/paginate.R
```diff
@@ -13,13 +13,20 @@
...
+#'   response. The result is passed to the argument `parsed` of `next_request()` and
+#'   `n_pages()`. This helps to avoid parsing the response multiple times.
+#' @param n_pages A function that extracts the total number of pages. It has two
+#'   arguments:
```
hadley, line 26:
```suggestion
#' @param n_pages An optional function that extracts the total number of pages, improving the 
#'   automatically generated progress bar. It has two arguments:
```

R/paginate.R
```diff
@@ -32,23 +39,28 @@
 #' request("https://pokeapi.co/api/v2/pokemon") %>%
 #'   req_url_query(limit = page_size) %>%
 #'   req_paginate_next_url(
-#'     next_url = function(resp) resp_body_json(resp)[["next"]],
-#'     n_pages = function(resp) {
-#'       total <- resp_body_json(resp)$count
+#'     next_url = function(resp, parsed) parsed[["next"]],
+#'     parse_resp = resp_body_json,
```
hadley, line 43:
```suggestion
#'     parse_resp = resp_body_json,
#'     next_url = function(resp, parsed) parsed[["next"]],
```

R/paginate.R
```diff
@@ -13,13 +13,20 @@
...
-#' @param next_request A callback function that takes a two arguments (the
-#'   original request and the response) and returns:
+#' @param next_request A callback function that returns a [request] to the next
+#'   page or `NULL` if there is no next page. It takes a three arguments:
 #'
-#'   * a new [request] to request the next page or
-#'   * `NULL` if there is no next page.
-#' @param n_pages A function that extracts the total number of pages from
-#'   the [response].
+#'   1. `req`: the original request.
```
hadley, line 19:
```suggestion
#'   1. `req`: the previous request.
```

R/paginate.R
```diff
@@ -32,23 +39,28 @@
...
 req_paginate <- function(req,
                          next_request,
+                         parse_resp = NULL,
                          n_pages = NULL) {
   check_request(req)
-  check_function2(next_request, args = c("req", "resp"))
-  check_function2(n_pages, args = "resp", allow_null = TRUE)
+  check_function2(next_request, args = c("req", "resp", "parsed"))
+  check_function2(parse_resp, args = "resp", allow_null = TRUE)
+  parse_resp <- parse_resp %||% function(resp) resp
```
hadley, line 56:
```suggestion
  parse_resp <- parse_resp %||% identity
```

R/paginate.R
```diff
@@ -58,10 +70,12 @@ req_paginate <- function(req,
 #'
 #' @inheritParams req_perform
 #' @param resp An HTTP [response].
+#' @param parsed The response parsed by the argument `parse_resp` of [req_paginate()].
 #' @param max_pages The maximum number of pages to request.
 #' @param progress Display a progress bar?
 #'
-#' @return A list of responses.
+#' @return A list of responses parsed with the `parse_resp` argument of
+#'   [req_paginate()].
```
hadley, line 78:
```suggestion
#'   [req_paginate()]. If this argument is not specified, it will be a list of responses.
```

Which I don't love because it makes the interface a bit type-unstable; but I don't see how to do better.

R/paginate.R
```diff
@@ -87,16 +102,19 @@ paginate_req_perform <- function(req,
   check_bool(progress)
 
   resp <- req_perform(req)
-  f_n_pages <- req$policies$paginate$n_pages %||% function(resp) Inf
-  n_pages <- min(f_n_pages(resp), max_pages)
+  parsed <- paginate_parse_response(resp, req)
+
+  f_n_pages <- req$policies$paginate$n_pages %||% function(resp, parsed) Inf
```
hadley, line 107: Do this in `req_paginate()` so we're setting all the default functions in one place?

R/utils.R
```diff
@@ -215,25 +215,18 @@ check_function2 <- function(x,
...
-      arg = arg
-    )
-  }
-
   if (n_actual_args == 0) {
-    arg_info <- "instead of no arguments"
+    arg_info <- "instead it has no arguments"
   } else {
-    arg_info <- "not {.arg {actual_args}}"
+    arg_info <- paste0("instead it has {.arg {actual_args}}")
```
hadley, line 229:
```suggestion
    arg_info <- paste0("it currently has {.arg {actual_args}}")
```
?

tests/testthat/_snaps/paginate.md
```diff
@@ -14,7 +14,7 @@
       req_paginate(req, function(req) req)
     Condition
       Error in `req_paginate()`:
-      ! `next_request` must have the arguments `req` and `resp`, not `req`.
+      ! `next_request` must have the arguments `req`, `resp`, and `parsed`, instead it has `req`.
```
hadley, line 17: Might be cleaer to use a semicolon to separate the clauses here.

hadley (approved):

R/paginate.R
mgirlich: At the moment everything errors and you get a not so helpful error message. Let's tackle this together with #310.

PR #302, opened 2023-09-04
Mocking enhancements

### LINKED ISSUE #301: Rename with_mock()
To something more specific, maybe `with_mocked_requests()`

And add `mock_requests_sequence` helper. 

Then improve oauth test coverage

hadley:
Fixes #301.

@mgirlich I think `local_mocked_responses()` is a better name, but `mocked_response_sequence()` seems long, especially since it'll often be used inside `local_mocked_responses()`. Any other ideas?


mgirlich:
The easiest would be to allow to directly pass a list to `with/local_mocked_responses()`. Either via a mutually exclusive argument `responses` or by allowing `mock` to be a list. Then apply the logic of `mocked_response_sequence()` on this list.

tests/testthat/test-paginate.R
```diff
@@ -43,30 +43,16 @@ test_that("req_paginate_next_url() checks inputs", {
...
-    }
-
-    body <- glue::glue(
-      '{"count":1281,"next":"<link>","previous":null,"results":[{"name":"bulbasaur","url":"https://pokeapi.co/api/v2/pokemon/1/"}]}',
-      .open = "<",
-      .close = ">"
-    )
-
-    response(
+  local_mocked_responses(mocked_response_sequence(
```
mgirlich, line 46: That is a very nice improvement 👍

tests/testthat/test-paginate.R
```diff
@@ -124,23 +110,15 @@ test_that("req_paginate_token() checks inputs", {
...
-
-    response(
-      url = req$url,
-      body = charToRaw(body),
-      headers = list(`content-type` = "application/json; charset=utf-8")
-    )
-  }
-  local_mock(next_url_mock)
+  local_mocked_responses(mocked_response_sequence(
+    response_json(body = list(x = 1, my_next_token = 2)),
```
mgirlich, line 114: Good catch! `response_json()` makes this so much nice! 😄

mgirlich (approved):

hadley:
Yeah, I like that idea.

R/resp.R
```diff
@@ -81,7 +81,7 @@ new_response <- function(method,
   headers <- as_headers(headers, error_call = error_call)
   # ensure we always have a date field
   if (!"date" %in% tolower(names(headers))) {
-    headers$Date <- http_date()
```
hadley, line 84: @mgirlich FYI. I think it's the right decision to do it here since `response()` is overwhelming used in tests/mocking.

PR #306, opened 2023-09-05
Add `paginate_req_perform_continue()`

mgirlich:
Closes #298.

``` r
devtools::load_all("~/GitHub/httr2/")
#> ℹ Loading httr2
page_size <- 7

req_pokemon <- request("https://pokeapi.co/api/v2/pokemon") %>%
  req_url_query(limit = page_size) %>%
  req_verbose(header_resp = FALSE) %>% 
  req_paginate_next_url(
    next_url = function(resp) resp_body_json(resp)[["next"]],
    n_pages = function(resp) {
      total <- resp_body_json(resp)$count
      ceiling(total / page_size)
    }
  )

responses <- paginate_req_perform(req_pokemon, max_pages = 2)
#> -> GET /api/v2/pokemon?limit=7 HTTP/2
#> -> Host: pokeapi.co
#> -> User-Agent: httr2/0.2.3.9000 r-curl/5.0.2 libcurl/8.1.2
#> -> Accept: */*
#> -> Accept-Encoding: deflate, gzip
#> -> 
#> -> GET /api/v2/pokemon?offset=7&limit=7 HTTP/2
#> -> Host: pokeapi.co
#> -> User-Agent: httr2/0.2.3.9000 r-curl/5.0.2 libcurl/8.1.2
#> -> Accept: */*
#> -> Accept-Encoding: deflate, gzip
#> ->

# oh, I actually wanted 3 pages, not 2
responses_all <- paginate_req_perform_continue(max_pages = 3)
#> -> GET /api/v2/pokemon?offset=14&limit=7 HTTP/2
#> -> Host: pokeapi.co
#> -> User-Agent: httr2/0.2.3.9000 r-curl/5.0.2 libcurl/8.1.2
#> -> Accept: */*
#> -> Accept-Encoding: deflate, gzip
#> ->
```

<sup>Created on 2023-09-05 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

PR #307, opened 2023-09-05
Add `req_chunk()`

### LINKED ISSUE #303: Add helper to perform a request in chunks
I frequently use such a helper when working with REST APIs:

* to insert/update/delete many objects
* to get the information on many objects where the endpoint actually uses `POST` as it uses a body

Usually, there is a limit on the number of elements that can be updated/requested, so the requests have to be chunked.
This is kind of similar to `req_paginate()` but the user has to do the "pagination" himself.

Below is an example of an implementation I use:

```r
api_perform_in_chunks <- function(req,
                                  x,
                                  f_body,
                                  chunk_size,
                                  progress = NULL,
                                  error_call = caller_env()) {
  chunks <- vec_chop_by_size(x, chunk_size)
  with_indexed_errors(
    purrr::map(
      chunks,
      \(chunk) {
        req <- httr2::req_body_raw(
          req,
          jsonlite::toJSON(f_body(chunk))
        )
        api_perform(req)
      },
      .progress = progress
    ),
    message = "In chunk {.field {cnd$location}}.",
    error_call = error_call
  )
}

vec_chop_by_size <- function(x, size) {
  n <- vctrs::vec_size(x)
  sizes <- vctrs::vec_rep(size, n %/% size)
  if (n %% size != 0) {
    sizes <- c(sizes, n %% size)
  }
  sizes <- vctrs::vec_cast(sizes, integer())

  vec_partition(x, sizes)
}

vec_partition <- function(x, sizes) {
  size <- vctrs::vec_size(x)

  if (size != sum(sizes)) {
    abort("`sizes` must add up to size of `x`.")
  }

  starts <- c(1L, sizes)
  starts <- cumsum(starts)[-length(starts)]
  starts <- starts - 1L # 0-based

  vctrs:::vec_chop_seq(x, starts, sizes)
}
```

mgirlich:
Closes #303.


mgirlich:
@hadley It would also make sense to add support for `paginate_req_perform_continue()` (see #306). Should we also add low level helpers? As this is pretty trivial I wasn't sure whether low level helpers are really worth it

R/chunk_req_perform.R
```diff
@@ -0,0 +1,79 @@
...
+#'       parsed <- resp_body_json(resp)
+#'       data.frame(
+#'         id = sapply(parsed, function(x) x$id),
+#'         name = sapply(parsed, function(x) x$name)
+#'       )
+#'     }
+#'   )
+#'
+#' responses
+chunk_req_perform <- function(req,
```
mgirlich, line 38: @hadley I think it might be better if we switch to the suffix form, i.e. `req_perform_chunk()`, `req_perform_paginate()` and `req_perform_multi()`. This makes it easier to discover by autocomplete.

hadley: Yeah, I was thinking this as well.

R/chunk_req_perform.R
```diff
@@ -0,0 +1,204 @@
+#' Chunk a request
+#'
```
hadley, line 2: Needs a description

R/chunk_req_perform.R
```diff
@@ -0,0 +1,204 @@
...
+#' }
+#'
+#' # in most cases `chunk_req_perform()` should be sufficient
+#' \dontrun{
+#' responses <- chunk_req_perform(
+#'   req,
+#'   chunk_size = 10,
+#'   data = ids,
+#'   apply_chunk = apply_chunk,
+#'   parse_resp = parse_resp
```
hadley, line 41: I wonder if it's worth adding a `combine_resp` argument too? (which would be `vctrs::vec_rbind()` or similar here). And that makes me wonder if it should be `resp_parse()` and `resp_combine()`.  And then should those become a top-level `req_parse()` that could be used by all the multi-request functions? Then maybe we have `req_perform_XYZ()` for returning lists of responses and `req_{something}_XYZ` for using the parsers?

R/chunk_req_perform.R
```diff
@@ -0,0 +1,204 @@
+#' Chunk a request
+#'
+#' @inheritParams req_perform
+#' @param chunk_size The size of each chunk.
+#' @param data The data to chunk.
```
hadley, line 5: Must be a vector?

R/chunk_req_perform.R
```diff
@@ -0,0 +1,204 @@
...
+  the$last_chunked_responses[seq2(1, the$last_chunk_idx)]
+}
+
+#' @export
+#' @rdname last_response
+last_chunk <- function() {
+  the$last_chunks[[the$last_chunk_idx]]
+}
+
+vec_chop_by_size <- function(x,
```
hadley, line 161: It feels a bit weird to bring in vctrs just for this function. But OTOH it would be a lot of work to accurately recreate the code in this package.

R/chunk_req_perform.R
```diff
@@ -0,0 +1,204 @@
...
+                      data,
+                      chunk_size,
+                      apply_chunk) {
+  check_request(req)
+  chunks <- vec_chop_by_size(data, chunk_size)
+  check_function2(apply_chunk, args = c("req", "chunk"))
+
+  n <- length(chunks)
+  requests <- lapply(chunks, function(chunk) apply_chunk(req, chunk))
+  new_chunked_request(requests, chunks)
```
hadley, line 77: This is different to all the other `req_` functions in that it now generates a list of requests. I think we need to either do the chunking lazily, not export this function, or give it a different prefix.

R/chunk_req_perform.R
```diff
@@ -0,0 +1,204 @@
...
+#' # ... which you can perform with `req_perform()`
+#' \dontrun{
+#' n <- length(requests)
+#' paths <- character(n)
+#' responses <- list()
+#' for (i in seq2(1, n)) {
+#'   paths[[i]] <- tempfile()
+#'   responses[[i]] <- req_perform(requests[[i]], path = paths[[i]])
+#' }
+#' }
```
hadley, line 66: It's interesting to think about using `multi_req_perform()` here.

R/chunk_req_perform.R
```diff
@@ -0,0 +1,204 @@
...
+  check_function2(parse_resp, args = "resp", call = error_call)
+
+  n <- length(requests)
+  the$last_chunked_responses <- vector("list", n)
+
+  pb <- create_progress_bar(total = n, name = "Request chunks", progress)
+  show_progress <- !is.null(pb)
+  env <- current_env()
+
+  for (i in seq2(1, n)) {
```
hadley, line 114: It feels like this should share more implementation with `multi_req_perform()`. Maybe we should expose the `host_con` argument from `curl::multi_set()`?

R/chunk_req_perform.R
mgirlich: I like the idea. One issue I see is that we basically need three functions:

1. `resp_parse(resp)`: parse the response. The result can be used in `n_pages` and `next_request`.
2. (new) `data(parsed)`: extract the actual data.
3. (new) `combine(data_list)`: combine the data.

An example where 3 functions are required is when the response body looks like `list(nextToken = ..., data = ...)`.

Maybe `resp_parse()` should always return a list where one field is `data`. This way we can avoid adding the 2nd function?

R/chunk_req_perform.R
mgirlich: I now also implemented this as a policy. This simplifies the interface of `chunk_req_perform()` and makes the whole API more similar to the pagination API.

R/chunk_req_perform.R
hadley: I think this is orthogonal to that problem, since I'm thinking more of just using `multi_req_perform()` here since you have a list of requests and vector of paths.

NEWS.md
```diff
@@ -9,6 +12,7 @@
 
 * `req_body_json()` gains custom content `type` argument
   and respects custom content-type set in header (@mgirlich, #189).
+>>>>>>> 62cd461e68e2d23de429d5f703a8034843d27a1c
```
hadley, line 15: Merge conflict

R/chunk_req_perform.R
hadley: Maybe it's worth starting a new issue for this — I think we need to look holistically at all requests that might return multiple responses, i.e. pagination, multi, and chunking. 

Maybe we need an interface more like `req_multi_chunk`, `req_multi_paginate` etc, and then you call something like `req_perform_multi` for all of them. (This function would have to be somewhat different to the current `multi_req_perform()` since it would need to handle pagination where you don't know all the pages up front, and we'd probably want to decouple from performing the requests in parallel).

hadley: Might be worth looking at python and rust libraries to see how they handle this problem.

mgirlich: I looked at a couple of python libraries and they don't have helpers for pagination or chunked requests.
For multiple requests in parallel they simply use the async capabilities (or in some cases threads) that python offers.

R/req-perform-multi.R
```diff
@@ -0,0 +1,227 @@
...
+#'   ids,
+#'   chunk_size = 3,
+#'   apply_chunk = apply_chunk
+#' )
+#'
+#' req1 <- multi_next_request(req, parsed = NULL)
+#' req1$url
+#' req2 <- multi_next_request(req1, parsed = NULL)
+#' req2$url
+multi_next_request <- function(req, parsed) {
```
mgirlich, line 208: I don't really like the name. Maybe `req_multi_next()`? Or only `req_next()`?

R/req-perform-multi.R
```diff
@@ -0,0 +1,228 @@
...
+#'   )
+#'
+#' responses <- req_perform_multi(req_pokemon)
+req_perform_multi <- function(req,
+                              # paths = NULL,
+                              max_requests = NULL,
+                              cancel_on_error = FALSE,
+                              progress = TRUE,
+                              error_call = current_env()) {
+  # TODO support `path` argument
```
mgirlich, line 49: Still working on this...

mgirlich:
@hadley Now `req_chunk()` and `req_paginate()` both use the new "multi" policy and can both be performed via the new `req_perform_multi()`. Overall, I think this goes in the right direction now.
What do you think of this now?

R/req-chunk.R
```diff
@@ -0,0 +1,100 @@
...
+#'   base_req,
+#'   chunk_size = 3,
+#'   data = ids,
+#'   apply_chunk = apply_chunk,
+#'   parse_resp = parse_resp
+#' )
+#'
+#' responses <- req_perform_multi(req)
+#' }
+req_chunk <- function(req,
```
mgirlich, line 51: It might be nice to have a helper to create a list of all requests so that one can perform them in parallel.
This would also work for two pagination patterns (pagination via offset and via page index). But this should be handled in a separate PR.

R/req-perform-multi.R
```diff
@@ -0,0 +1,228 @@
...
+  )
+
+  parse_resp <- error_wrapper(
+    f = function(resp) {
+      parse_resp <- req$policies$parse_resp
+      parse_resp(resp)
+    },
+    cancel_on_error = cancel_on_error,
+    error_message = "When parsing response {i}.",
+    error_class = "httr2_parse_failure",
```
mgirlich, line 82: Or would you prefer to have the same class when the request fails and when the parsing fails? I.e. <httr2_failure> in both cases.

R/req-perform-multi.R
```diff
@@ -0,0 +1,228 @@
...
+    }
+  }
+  if (show_progress) cli::cli_progress_done()
+
+  # remove unused tail of `out`
+  if (i < length(out)) {
+    out <- out[seq2(1, i)]
+  }
+
+  vctrs::list_unchop(out)
```
mgirlich, line 121: This doesn't work well in case `cancel_on_error = FALSE` and there actually is an error. Should we only unchop if `cancel_on_error = TRUE`?

R/req-perform-multi.R
```diff
@@ -0,0 +1,228 @@
...
+#'       list(data = data, next_url = parsed$`next`)
+#'     },
+#'     n_pages = function(parsed) {
+#'       total <- parsed$count
+#'       ceiling(total / page_size)
+#'     }
+#'   )
+#'
+#' responses <- req_perform_multi(req_pokemon)
+req_perform_multi <- function(req,
```
mgirlich, line 43: We need to think about the naming of functions. There currently is `multi_req_perform()` to perform a list of requests in parallel. It might make sense to rename this to something like `reqs_perform_parallel()` (though I'm not quite sure we want to add the prefix `reqs`).

R/req-chunk.R
```diff
@@ -0,0 +1,100 @@
+#' Chunk a request
```
mgirlich, line 1: Or should this be `Batch a request` and the function name `req_batch()`?

R/paginate.R
```diff
@@ -4,10 +4,8 @@
 #' `r lifecycle::badge("experimental")`
 #'
 #' Use `req_paginate()` to specify how to request the next page in a paginated
-#' API. Use [paginate_req_perform()] to fetch all pages.
 #' If you need more control use a combination of [req_perform()] and
-#' [paginate_next_request()] to iterate through the pages yourself.
-#'
+#' [req_next_multi()] to iterate through the pages yourself.
```
hadley, line 8: Or should this be `resp_next_request()`?

R/req-chunk.R
hadley: Yeah, I like batch better.  `req_` is probably not quite the right prefix (since it doesn't return a request) but I don't see a better alternative.

R/req-perform-multi.R
hadley: Or only unchop the results corresponding to requests that succeeded?

R/req-perform-multi.R
```diff
@@ -0,0 +1,245 @@
...
+                          env = caller_env(),
+                          error_call = caller_env()) {
+  force(env)
+
+  if (cancel_on_error) {
+    function(...) {
+      tryCatch(
+        f(...),
+        error = function(cnd) {
+          cli::cli_abort(
```
hadley, line 174: Fits on one line?

hadley (commented):
I'll do another review once we've sorted out #335 because I'm getting confused by the different names.

PR #315, opened 2023-09-08
Simplify `req_paginate()` helpers

mgirlich:
As written in https://github.com/r-lib/httr2/pull/307#discussion_r1318380338 I realised that we can simplify the overall interface when `parse_resp()` returns a list where one field must be `data`.
E.g. for `req_paginate_next_url()` we simply require another field `next_url` and can get rid of the argument `next_url`.
Also, we can now actually only return the data instead of accidentally also returning `next_url` and this makes it much easier to combine the data.

The pokemon api example is now a bit longer but directly extracts all the data:

```r
page_size <- 25

req_pokemon <- request("https://pokeapi.co/api/v2/pokemon") %>%
  req_url_query(limit = page_size) %>%
  req_paginate_next_url(
    parse_resp = function(resp) {
      parsed <- resp_body_json(resp)
      results <- parsed$results
      data <- tibble::tibble(
        name = purrr::map_chr(results, "name"),
        url = purrr::map_chr(results, "url"),
      )
      list(data = data, next_url = parsed$`next`)
    },
    n_pages = function(parsed) {
      total <- parsed$count
      ceiling(total / page_size)
    }
  )

responses <- paginate_req_perform(req_pokemon)
responses
```


R/paginate.R
```diff
@@ -155,31 +170,24 @@ paginate_req_perform <- function(req,
...
-                                  next_url,
-                                  parse_resp = NULL,
+                                  parse_resp,
                                   n_pages = NULL) {
-  check_function2(next_url, args = c("resp", "parsed"))
-
-  next_request <- function(req, resp, parsed) {
-    next_url <- next_url(resp, parsed)
+  next_request <- function(req, parsed) {
+    next_url <- parsed[["next_url"]]
```
mgirlich, line 190: The question is how/where to document it best, that `parse_resp()` needs to return a list that has field `next_url` for `req_paginate_next_url()`?

R/paginate.R
```diff
@@ -135,7 +150,7 @@ paginate_req_perform <- function(req,
     resp <- req_perform(req)
     parsed <- parse_resp(resp)
 
-    out[[page]] <- parsed
+    out[[page]] <- parsed$data
```
hadley, line 153: If we do this, I wonder if we should `vec_c()` the results together at the end. That will work for most "data" types in R, and if it doesn't work for you, you can always wrap in a `list()`.

mgirlich: I like the idea and added this. It adds a dependency on vctrs but I think that vctrs is so commonly used anyways that it doesn't really matter much.

R/paginate.R
```diff
@@ -14,14 +14,22 @@
 #'
 #' @inheritParams req_perform
 #' @param next_request A callback function that returns a [request] to the next
-#'   page or `NULL` if there is no next page. It takes a three arguments:
+#'   page or `NULL` if there is no next page. It takes a two arguments:
```
hadley, line 17:
```suggestion
#'   page or `NULL` if there is no next page. It takes two arguments:
```

R/paginate.R
```diff
@@ -39,29 +47,55 @@
...
-  n_pages <- n_pages %||% function(resp, parsed) Inf
+  parse_resp <- parse_resp %||% function(resp) list(data = resp)
+  check_character(required_fields, allow_null = TRUE)
+  required_fields <- union(required_fields, "data")
+  check_function2(n_pages, args = "parsed", allow_null = TRUE)
+  n_pages <- n_pages %||% function(parsed) Inf
+
+  wrapped_parse_resp <- function(resp) {
+    out <- parse_resp(resp)
+    if (!is.list(out)) {
```
hadley, line 81: Since we're taking a dependency on vctrs, I think this could now be `vctrs::obj_check_list()`

R/paginate.R
```diff
@@ -70,14 +104,14 @@ req_paginate <- function(req,
...
 #' @param parsed The response parsed by the argument `parse_resp` of [req_paginate()].
 #' @param max_pages The maximum number of pages to request.
 #' @param progress Display a progress bar? Use `TRUE` to turn on a basic progress
 #'   bar, use a string to give it a name, or see [progress_bars] for more details.
 #'
-#' @return A list of responses parsed with the `parse_resp` argument of
-#'   [req_paginate()]. If this argument is not specified, it will be a list of responses.
+#' @return The result of `vec_c()`ing together the `data` fields that were
+#'   extracted by the `parse_resp()` argument of [req_paginate()].
+#'   If this argument is not specified, it will be a list of the raw responses.
```
hadley, line 114:
```suggestion
#'   If `parse_resp` is not specified, it will be a list of the raw responses.
```

R/paginate.R
```diff
@@ -70,14 +104,14 @@ req_paginate <- function(req,
...
-#' @param resp An HTTP [response].
 #' @param parsed The response parsed by the argument `parse_resp` of [req_paginate()].
 #' @param max_pages The maximum number of pages to request.
 #' @param progress Display a progress bar? Use `TRUE` to turn on a basic progress
 #'   bar, use a string to give it a name, or see [progress_bars] for more details.
 #'
-#' @return A list of responses parsed with the `parse_resp` argument of
-#'   [req_paginate()]. If this argument is not specified, it will be a list of responses.
+#' @return The result of `vec_c()`ing together the `data` fields that were
+#'   extracted by the `parse_resp()` argument of [req_paginate()].
```
hadley, line 113:
```suggestion
#'   extracted by the `parse_resp` argument of [req_paginate()].
```

hadley (approved):

PR #318, opened 2023-09-15
Fix unit tests on Workbench

atheriel:
Previously these tests would fail because `is_hosted_session()` returns `TRUE` on these platforms and yields a different error. This commit skips these tests in hosted sessions instead.

### DIFF
--- a/tests/testthat/test-oauth-flow-auth-code.R
+++ b/tests/testthat/test-oauth-flow-auth-code.R
@@ -51,11 +51,17 @@ test_that("bare authorisation codes can be input manually", {
 # normalize_redirect_uri --------------------------------------------------
 
 test_that("adds port to localhost url", {
+  # Allow tests to run when is_hosted_session() is TRUE.
+  local_mocked_bindings(is_hosted_session = function() FALSE)
+
   redirect <- normalize_redirect_uri("http://localhost")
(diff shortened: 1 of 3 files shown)


tests/testthat/test-oauth-flow-auth-code.R
```diff
@@ -51,16 +51,20 @@ test_that("bare authorisation codes can be input manually", {
 # normalize_redirect_uri --------------------------------------------------
 
 test_that("adds port to localhost url", {
+  skip_if(is_hosted_session(), "Skipping in hosted sessions")
```
hadley, line 54: I think it would be slightly better just to mock these out, i.e.

```R
local_mocked_bindings(is_hosted_session = function() FALSE)
```

And then if you were really keen you could add a specific test for the error when it's `TRUE`. (But if you don't have time I can do it too)

atheriel: Done.

tests/testthat/test-oauth-flow-auth-code.R
```diff
@@ -72,6 +78,14 @@ test_that("old args are deprecated", {
 
 })
 
+test_that("hosted sessions error on localhost redirects", {
```
hadley: Doh we already had a test for this 😬

PR #320, opened 2023-09-25
Add support for using Workbench's generic OAuth2 code capture

atheriel:
For data scientists using local tools or working in the IDE, authentication with APIs or databases that use the OAuth2 authorisation code flow (e.g. Google, Azure, or GitHub) typically work by running a temporary HTTP server on `localhost` and using that as the redirect URL. When a user gets redirected to `localhost`, the temporary HTTP server captures the code automatically and surfaces it to the caller.

Unfortunately, this doesn't work at all on Workbench, because you can't just "redirect to `localhost`" like you can on your desktop.

In #248 and follow-ups we added a workaround for this issue that we term the "pseudo out-of-band" flow -- that mechanism was pioneered by the `gargle` package in response to Google deprecating the original "out-of-band" flow.

This commit adds support for a *third* mechanism for the auth code flow which uses a feature of the upcoming Workbench release that allows it to serve as a redirect URL for arbitrary OAuth2 applications on a static `/oauth_redirect_callback` endpoint. `httr2` can retrieve any code sent to this URL by calling a simple JSON API at a static `/oauth_code` endpoint.

This mechanism has strong appeal over the pseudo out-of-band flow because it sidesteps the requirement that users copy & paste the code, making it feel much more natural and automatic. It also avoids the need to ask users to host a static "capture your code" page, a la https://www.tidyverse.org/google-callback/, because every Workbench instance now has one.

(This Workbench feature was explicitly designed for packages like `httr2` to make use of so that the auth code flow starts feeling like magic once again.)

I've decided to introduce client-specific environment variables here -- specifically, `HTTR2_OAUTH_REDIRECT_URL` and `HTTR2_OAUTH_CODE_SOURCE_URL` -- rather than having platform-specific ones prefixed with `WORKBENCH_`.

If you're running a daily build of Workbench locally, you can test this as follows:

```r
local({
  # Automatically determine the Workbench server's URL. We use
  # RSTUDIO_HTTP_REFERER here, which is only set in RStudio Pro sessions.
  url_split <- strsplit(Sys.getenv("RSTUDIO_HTTP_REFERER"), "/s/", fixed = TRUE)[[1]]
  if (nchar(url_split[1]) != 0L && length(url_split) == 2L) {
    base_url <- url_split[1]
    Sys.setenv(
      HTTR2_OAUTH_REDIRECT_URL = sprintf("%s/oauth_redirect_callback", base_url),
      HTTR2_OAUTH_CODE_SOURCE_URL = sprintf("%s/oauth_code", base_url)
    )
  }
})

client <- oauth_client(
  id = "28acfec0674bb3da9f38",
  secret = obfuscated(paste0(
    "J9iiGmyelHltyxqrHXW41ZZPZamyUNxSX1_uKnv",
    "PeinhhxET_7FfUs2X0LLKotXY2bpgOMoHRCo"
  )),
  token_url = "https://github.com/login/oauth/access_token",
  name = "hadley-oauth-test"
)

oauth_flow_auth_code(
  client, auth_url = "https://github.com/login/oauth/authorize"
)
```

In the future we can set the environment variables in Workbench sessions directly.

Unit tests are included.


R/oauth-flow-auth-code.R
```diff
@@ -440,3 +446,33 @@ oauth_flow_auth_code_read <- function(state) {
 read_line <- function(prompt = "") {
   readline(prompt)
 }
+
+# base::readline() wrapper so we can mock user input during testing.
+read_line <- function(prompt = "") {
```
hadley, line 451: Dupe?

R/oauth-flow-auth-code.R
```diff
@@ -59,7 +59,7 @@ req_oauth_auth_code <- function(req, client,
                                 pkce = TRUE,
                                 auth_params = list(),
                                 token_params = list(),
-                                redirect_uri = "http://localhost",
+                                redirect_uri = Sys.getenv("HTTR2_OAUTH_REDIRECT_URL", "http://localhost"),
```
hadley, line 62: I'd suggest making this an exported helper function so you can document how the default is computed.

atheriel: Like `default_redirect_uri()`?

hadley: Yeah, exactly.

atheriel: Done.

R/oauth-flow-auth-code.R
atheriel: Fixed.

R/oauth-flow-auth-code.R
```diff
@@ -201,7 +212,12 @@ oauth_flow_auth_code <- function(client,
   )
   utils::browseURL(user_url)
 
-  if (redirect$localhost) {
+  if (redirect$can_fetch_code) {
+    # Wait a bit to give the user a chance to click through the authorisation
+    # process.
+    sys_sleep(2, "for browser-based authentication", progress = FALSE)
```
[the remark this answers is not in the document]
atheriel: I'm uncertain how to handle this. I really don't love that we block here, it sounds pretty annoying for e.g. Shiny apps. But I also think it would be OK to punt this to a follow-up PR. What do you think?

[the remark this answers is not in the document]
hadley: I don't have strong feelings. We can always make it better later.

R/oauth-flow-auth-code.R
```diff
@@ -201,7 +212,15 @@ oauth_flow_auth_code <- function(client,
   )
   utils::browseURL(user_url)
 
-  if (redirect$localhost) {
+  if (redirect$can_fetch_code) {
+    # Wait a bit to give the user a chance to click through the authorisation
+    # process.
+    if (!is_testing()) {
```
hadley, line 218: The fact that we need to do this to make the tests faster makes me think polling would be the better way to go.

tests/testthat/test-oauth-flow-auth-code.R
```diff
@@ -90,5 +90,51 @@ test_that("forwards oauth error", {
...
+    expect_false(can_fetch_oauth_code("http://localhost:9090/redirect"))
+  })
+})
+
+# oauth_flow_auth_code_fetch ----------------------------------------------
+
+test_that("auth codes can be retrieved from an external source", {
+  skip_on_cran()
+
+  app <- webfakes::new_app()
```
hadley, line 119: This test is a bit easier to write with webfakes (which we already use elsewhere)

PR #326, opened 2023-09-29
Enable `oauth_flow_auth_code` to parse URL for component `code` and `state`

fh-mthomson:
If a `redirect_uri` is a static redirect i.e., does _not_ parse results via Javascript (as done in https://www.tidyverse.org/google-callback ([parsing source code](https://github.com/r-lib/gargle/blob/main/inst/pseudo-oob/google-callback/index.html)), users currently need to manually parse `code` and `state` from the URL itself, which is time consuming (and error prone). 

Rather, it would be more efficient to just copy the URL itself, rather than require users to host a site that parses the result.

@hadley this suggested change is based on our convo during the Posit conf workshop!


hadley:
Can you explain a bit more about how you expect the user experience to work? You'll forward them to a page that just says "copy this url"? If you're hosting that page, why is it any harder to copy in the javascript that would parse the components for them?

fh-mthomson:
> You'll forward them to a page that just says "copy this url"?

Yeah, exactly!

> If you're hosting that page, why is it any harder to copy in the javascript that would parse the components for them?

It's a _very_ static page, i.e.,: `https://gitlab.com/<internalrepo>/-/blob/main/readme.md`

So, it would be ideal to not have to create javascript (our team doesn't have immediate expertise) and instead defer to the `httr2` workflows to handle all OAuth parsing workflows. 

The current alternative is to have folks manually extract `code` and `state` entries from the URL.

hadley:
You don’t have to create JavaScript, you can just copy it onto that page?

calkunz:
To clarify Michael's comment, 

> It's a _very_ static page, i.e.,: `https://gitlab.com/<internalrepo>/-/blob/main/readme.md`

For our redirect, the Gitlab markdown file we're using is a static page that's essentially a page of documentation within a larger repo, so it isn't a full site where we are able to interact with the JavaScript and have it handle the parsing

hadley:
Ah, it's not even an HTML page.

fh-mthomson:
thank you, @hadley !

Issue #120, opened 2022-04-01
`req_oauth_device` doesn't use `auth_params`


jl5000:
I'm trying to pass a `auth_url` via `auth_params` in the `req_oauth_device` flow and it seems to be giving an error message of `argument "auth_url" is missing, with no default ` despite me passing it. I've stepped through the code and it seems to be getting lost during the `exec()` call in `auth_oauth_sign()`.

I'm using it like this:

```
req %>% 
    httr2::req_oauth_device(cl, 
                            auth_params = list(auth_url = "http://xyz/oauth2/authorize"),
                            cache_disk = TRUE) %>% 
    httr2::req_perform()
```

hadley:
Because `auth_url` is its own argument:

```R
req %>%
  httr2::req_oauth_device(
    client = cl,
    auth_url = "http://xyz/oauth2/authorize",
    cache_disk = TRUE
  ) %>%
  httr2::req_perform()
```

PR #331, opened 2023-10-06
Pass auth_url param through to oauth_flow_device

taerwin:
auth_url is not passed through to oauth_device_flow see issue https://github.com/r-lib/httr2/issues/120


hadley:
Could you please explain why you want this?

taerwin:
There is currently no way to set auth_url in req_oath_device() directly or via auth_params. Ticket suggests it should be able to be set as follows:

req %>%
  httr2::req_oauth_device(
    client = cl,
    auth_url = "http://xyz/oauth2/authorize",
    cache_disk = TRUE
  ) %>%
  httr2::req_perform()

This patch adds auth_url as a param and also sets this param to be passed through to oauth_flow_device

hadley:
Ah, got it.

PR #335, opened 2023-10-09
Use common `req_perform_` prefix

### LINKED ISSUE #314: Rename `paginate_req_perform()` to `req_perform_paginate()` etc
This makes it much easier to discover by autocomplete.

hadley:
Fixes #314


R/multi-req.R
```diff
@@ -44,19 +44,19 @@
...
 #' )
-#' # multi_req_perform() will always succeed
-#' resps <- multi_req_perform(reqs)
+#' # req_perform_multi() will always succeed
+#' resps <- req_perform_multi(reqs)
 #' # you'll need to inspect the results to figure out which requests fails
 #' fail <- vapply(resps, inherits, "error", FUN.VALUE = logical(1))
 #' resps[fail]
-multi_req_perform <- function(reqs, paths = NULL, pool = NULL, cancel_on_error = FALSE) {
+req_perform_multi <- function(reqs, paths = NULL, pool = NULL, cancel_on_error = FALSE) {
```
mgirlich, line 59: Maybe this should be `req_perform_parallel()`? At least this sounds a bit more intuitive to me.

mgirlich:
I'm not sure we want to stick with `req_perform_paginate()` in the end. If we add `req_chunk()` we would either have to add `req_perform_chunk()` or change `req_chunk()` to return a list of requests.

R/multi-req.R
hadley: Oh yeah, I like that name.

hadley:
How about `req_perform_iterate()`?

hadley:
@mgirlich should we rename `req_stream()` to `req_perform_stream()`?

mgirlich:
I like `req_perform_stream()` and `req_perform_iterate()`!

R/multi-req.R
```diff
@@ -19,7 +19,7 @@
 #' * Consults the cache set by [req_cache()] before/after all requests.
 #'
 #' In general, where [req_perform()] might make multiple requests due to retries
-#' or OAuth failures, `multi_req_perform()` will make only make 1.
+#' or OAuth failures, `req_perform_parallel()` will make only make 1.
```
mgirlich, line 22:
```suggestion
#' or OAuth failures, `req_perform_parallel()` will only make 1.
```

R/paginate.R
```diff
@@ -110,15 +110,15 @@ req_paginate <- function(req,
   )
 }
 
-#' Perform a paginated request
+#' Perform a iterated request
```
mgirlich, line 113:
```suggestion
#' Perform an iterated request
```

R/paginate.R
```diff
@@ -110,15 +110,15 @@ req_paginate <- function(req,
...
 #'
 #' @description
 #' `r lifecycle::badge("experimental")`
 #'
-#' * `paginate_req_perform()` requests all pages for a paginated request and
+#' * `req_perform_iterate()` requests all pages for a iterated request and
```
mgirlich, line 118:
```suggestion
#' * `req_perform_iterate()` requests all pages for an iterated request and
```

R/paginate.R
```diff
@@ -110,15 +110,15 @@ req_paginate <- function(req,
...
 #'   returns a list of responses.
-#' * `paginate_next_request()` generates the request for the next page for a
-#'   paginated response, or `NULL` if there are no more pages to return.
+#' * `iterate_next_request()` generates the request for the next page for a
```
mgirlich, line 120:
```suggestion
#' * `iterate_next_request()` generates the request for the next page for an
```

mgirlich (approved):

hadley:
I really like this section now. Feels like the functions fit together into a whole.

Screenshot 2023-10-12 at 16 30 47

hadley:
This makes me lean towards making the next release 1.0.0 again.

hadley:
I'm swinging towards `req_perform_iteratively()` since that would match `req_perform_parallel()` better — i.e. perform iteratively; perform in parallel.

mgirlich:
`req_perform_iteratively()` also sounds good and I prefer it over `req_perform_sequential()`.

One use case that comes to my mind for `req_perform_sequential()` would be for requests that create something. E.g. consider GitHub: the first request creates a repository, the second request creates an issue in this repository, the third request creates a comment in this issue.

Harder to come up with something for GET requests. Basically the only things that comes to my mind is if you don't want to do too many requests (e.g. because they cost something) but you don't know whether you will need all of them.
But the most realistic case for something like this would probably be a paginated search endpoint, where you don't (really) know all the requests in advance.
This brings me to the idea that for this we might want to include a `break` callback to `req_perform_iteratively()` at some point.

hadley:
I'll make the change to `req_perform_iteratively()` and then merge this PR. Let's continue to talk about the interface for iterative requests; I think if we can get this right, I'll feel happy making the next release 1.0.0 and it's worth taking some extra time to do that.

PR #342, opened 2023-10-12
Separate out iterate_next_request() docs

### LINKED ISSUE #338: Example for `paginate_next_request()`
An example for `paginate_next_request()` would help in attempting to use

> a combination of req_perform() and paginate_next_request() to iterate through the pages yourself.

After digging in the test-paginate.R, I see this is the basic idea of `paginate_next_request()`:

```
req <- my_req() |> 
  httr2::req_paginate(
    next_request = function(req, parsed) {
      if(parsed$next_page == "") return(NULL)
      req |> req_url(parsed$next_page)
    }
  )

parsed <- req_perform(req) |> my_parse()
next_req <- req |> httr2::paginate_next_request(parsed = parsed)
```

Perhaps it is appropriate to include a full example of how one might perform the iteration, as per @hadley in https://github.com/r-lib/httr2/issues/8#issuecomment-866221516_:

```
`%||%` <- function(x, y) {
  if (is.null(x)) y else x
}

my_req_perform_paginate <- function(req, max_pages = NULL) {
  out <- vector("list", max_pages %||% 10)
  i <- 1L
  
  repeat({
    out[[i]] <- req_perform(req) |> mm_parse()
    if (!is.null(max_pages) && i == max_pages) {
      break
    }
    
    req <- req |> httr2::paginate_next_request(parsed = out[[i]])
    
    if (is.null(req)) {
      break
    }
    
    i <- i + 1L
    if (i > length(out)) {
      length(out) <- length(out) * 2L
    }
  })
  
  if (i != length(out)) {
    out <- out[seq_len(i)]
  }
  out
  
}
```

hadley:
And add example. Fixes #338


R/paginate.R
```diff
@@ -215,8 +212,31 @@ req_perform_iteratively <- function(req,
...
+#' @examples
+#' req_flowers <- request(example_url()) %>%
+#'   req_url_path("/iris") %>%
+#'   req_url_query(limit = 40) %>%
+#'   req_paginate_page_index(
+#'     page_index = function(req, page) {
+#'       req %>% req_url_query(page_index = page)
+#'     }
+#'   )
+#' req_flowers %>% .$url
```
mgirlich, line 235: The pipe seems a bit unnecessary?

```suggestion
#' req_flowers$url
```

R/paginate.R
```diff
@@ -215,8 +212,31 @@ req_perform_iteratively <- function(req,
...
+#'
+#' resp <- req_flowers %>% req_perform()
+#' next_req <- iterate_next_request(req_flowers, resp)
+#' next_req %>% .$url
```
mgirlich, line 239:
```suggestion
#' next_req$url
```

R/paginate.R
```diff
@@ -215,8 +212,31 @@ req_perform_iteratively <- function(req,
   vctrs::list_unchop(out)
 }
 
+#' Retrieve the next request from an iterative response
+#'
+#' In most case you should use [req_perform_iteratively()] but you can
+#' use this lower-level helper if you need more control.
```
mgirlich, line 218: This seems a bit too short. Maybe something like:

In most case you should use [req_perform_iteratively()] but you can use this lower-level helper to iterate through the requests and perform them yourself with [req_perform()]

mgirlich (approved):

PR #349, opened 2023-10-16
Implement oauth_token_cached()

### LINKED ISSUE #328: Move token (a) caching and (b) refresh token layer from `req_oauth_*()` to `oauth_flow*()`
**Goal**: use `oauth_flow_auth_code()` as a single function through which I can:

1. **Generate**: Create a token
2. **Cache**: Initialize/leverage a cache using ^ token (via `cache_choose()` et al)
3. **Refresh**: If a given token (new or cached) is expired, use the refresh token to generate a new token (rather than `oauth_flow_refresh()`, as discussed at length in #186)

Currently, only (1) is supported by `oauth_flow_auth_code()` whereas (2) & (3) are automated via [auth_outh_sign()](https://github.com/r-lib/httr2/blob/f9e73a3036f00dc96e499f9377b1e3a134b5ff8c/R/oauth.R#L31C1-L55) but only as an indirect side effect of calling `req_oauth_auth_code()`. 

**Proposal**: move the `auth_outh_sign()` logic to the token-generation layer, instead of (or in addition to) the request layer, since, e.g., `req_oauth_auth_code()` calls `oauth_flow_auth_code()` anyway.

hadley:
Fixes #328

@fh-mthomson does this look like it will be adequate for your needs? I think your code would look something like this:

```R
oauth_token <- function() {
  OAUTH_AUTHORIZE_ENDPOINT <- "https://<X>.okta.com/oauth2/<X>/v1/authorize"
  REDIRECT_URI <- "https://<X>"
  SCOPE <- "session:role-any"
  
  oauth_token_cached(
    client = client,
    flow = oauth_flow_auth_code,
    flow_params = list(
      auth_url = OAUTH_AUTHORIZE_ENDPOINT,
      scope = SCOPE,
      redirect_uri = REDIRECT_URI
    ),
    cache_disk = TRUE
  )
}

oauth_client <- function() {
  OAUTH_CLIENT_ID <- "<X>"
  OAUTH_TOKEN_ENDPOINT <- "https://<X>.okta.com/oauth2/<X>/v1/token"
  
  httr2::oauth_client(
    id = OAUTH_CLIENT_ID,
    token_url = OAUTH_TOKEN_ENDPOINT,
    name = "oauth_client"
  )
}
```


R/oauth.R
```diff
@@ -48,10 +57,56 @@ auth_oauth_sign <- function(req, reauth = FALSE) {
       } else {
         token <- token_refresh(flow_params$client, token$refresh_token)
```
fh-mthomson, line 58: Currently, I have a cached token with dually-expired values (after coming back from vacation):

1. `access_token` (expired after X hrs, <1 day)
2. `refresh_token` (expired after Y days, e.g., ~1 week)

Accordingly, `token_refresh()` code fails with the error:
```
Error in `token_refresh()`:
! OAuth failure [invalid_grant]
• The refresh token is invalid or expired.
```

Would it be reasonable to add a `tryCatch()` (or otherwise) such that if `token_refresh()` fails, per the cached `token$refresh`, then we fall back to `token <- exec(flow, !!!flow_params)`?

fh-mthomson:
This is exactly what I was hoping for! Thank you!

R/oauth.R
fh-mthomson: Looks like #168 shared by @jennybc handles `N` refreshes with a _valid_ refresh token. 

I'm curious if you've thought about how _expired_ refresh tokens are handled?

hadley: I have no thought about expired refresh tokens at all 😭 But I think it seem reasonable to try again here if `token_refresh()` fails with a `httr2_oauth` error.  Or maybe even `httr2_oauth_invalid_grant`?

Does the API you're using happen to document this somewhere I can read about it?

hadley: From your other issues, maybe https://developer.okta.com/docs/guides/refresh-tokens/main/#refresh-token-rotation ?

fh-mthomson: Yes, exactly! We are using refresh token rotation per those docs, with a set expiration of N days. So, we'd expect to `token_refresh()` to error on days ≥ N+1.

hadley: I also noticed from the google docs that this also applies there, although google token expires after 6 months of no use, and httr2 deletes cached tokens after 30 days.

fh-mthomson: nice! Thanks again for adding this @hadley, it's saved us a lot of code overhead and made the auth process ~trivial!

Issue #341, opened 2023-10-12
`iterate_next_request()` -> `resp_next_request()`


hadley:
This would mean changing the arguments so that it takes a request as the first argument, and maybe automatically attaching the `parsed` data and `next_request` call back to that request.

That possibly implies that registering a parser is a more general feature, which we could possibly implement lazily. i.e. you'd register a parser in the request (with `req_parser()`?), and then call `resp_parsed()` (or maybe `resp_body_parsed()`) to access the parsed body. We'd need to add an environment to the response so that we could do the parsing once and cache it. And then _maybe_ have some helper that does `resp_parse(req_perform(req))`? And something similar for `req_perform_parallel()`?  That would make `req_perform_iterate()` fit in better.

@mgirlich these are very quick and rough notes, but does this make sense to you?

hadley:
Actually, since we're introducing `req_peform_parallel()` we could make it automatically parse, just like `req_perform_iterate()`.

mgirlich:
I agree that it might make sense to add a separate `req_parser()`. The only thing I'm not quite sure about is documentation and return type.
For `req_paginate()` the output of `parse_resp()` should be a list with a field `data` and any other fields that are needed for the pagination to work. For `req_batch()` and `req_perform_parallel()` returning a list with a field `data` would feel artificial.

Automatically parsing and `vec_c()`ing the results in `req_peform_parallel()` and `req_perform_iteratively()` is awesome, so we should definitely do that.

> This would mean changing the arguments so that it takes a request as the first argument, and maybe automatically attaching the parsed data and next_request call back to that request.

It currently takes a request as the first argument. Did you mean response?

hadley:
Yeah, sorry, a response as a first argument

hadley:
Maybe we need `req_parse_data()` and `req_parse_metadata()`? Or two arguments to `req_parse()`? But that's not quite right for the case (like GitHub) where a paginated request returns a both data and metadata in the json. So maybe the `req_parse()` function should return a named list containing data and metadata? But then that feels weird for the straightforward `req_perform_parallel()` case. Hmmmm.

(Also I think it makes sense for `req_perform()` to automatically parse the result if you've used `req_parse()`. That feels a little type-unstable, but it's very convenient for the other perform functions and will be backward compatible).

hadley:
Maybe for `req_perform_iterative()` we require a `req_parse()` that provides that named list, but for others we don't?

hadley:
[...]

PR #351, opened 2023-10-17
Cache the result of `resp_body_json()`

hadley:
Part of #341

### DIFF
--- a/R/resp-body.R
+++ b/R/resp-body.R
@@ -3,7 +3,8 @@
 #' @description
 #' * `resp_body_raw()` returns the raw bytes.
 #' * `resp_body_string()` returns a UTF-8 string.
-#' * `resp_body_json()` returns parsed JSON.
+#' * `resp_body_json()` returns parsed JSON. The result is cached in the
+#'   response so that calling `resp_body_json()` repeatedly is cheap.
 #' * `resp_body_html()` returns parsed HTML.
 #' * `resp_body_xml()` returns parsed XML.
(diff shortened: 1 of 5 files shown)


R/resp-body.R
```diff
@@ -77,6 +78,10 @@ resp_body_string <- function(resp, encoding = NULL) {
 #' @rdname resp_body_raw
 #' @export
 resp_body_json <- function(resp, check_type = TRUE, simplifyVector = FALSE, ...) {
+  if (env_has(resp$cache, "json")) {
```
mgirlich, line 81: This completely ignores the arguments passed to `jsonlite::fromJSON()`. Usually, this shouldn't be an issue as it doesn't make much sense to use different argument. The exception is probably when requesting a new endpoint and you don't really know yet which arguments you want to use.

R/resp-body.R
```diff
@@ -107,6 +113,11 @@ resp_body_html <- function(resp, check_type = TRUE, ...) {
 #' @rdname resp_body_raw
 #' @export
 resp_body_xml <- function(resp, check_type = TRUE, ...) {
+  if (env_has(resp$cache, "xml")) {
```
mgirlich, line 116: Same thing regarding the arguments

mgirlich (approved):

R/resp-body.R
hadley: Ooops, good point. Let me add that to the cache key.

PR #352, opened 2023-10-17
Make it easier to generate repeated query parameters

### LINKED ISSUE #350: pass multi-value parameters as vectors to GET
GET requests can handle multi-value parameters, although there is [no defined standard](https://stackoverflow.com/a/24728298/2375551) on how to do it. `httr2` currently supports multi-value parameters when you define them explicitly as separate arguments:
```r
request("example.com") |>
  req_url_query(foo = 1, foo = 2) |>
  req_dry_run()

#> GET /?foo=1&foo=2 HTTP/1.1
#> Host: example.com?foo=1&foo=2
#> User-Agent: httr2/0.2.3 r-curl/5.1.0 libcurl/8.3.0
#> Accept: */*
#> Accept-Encoding: deflate, gzip
```

But it would be nice if you could pass multi-value parameters as vectors and have httr2 do the work of forming the url with multiple entries for the parameter(s):

```r
# desired  behaviour
request("example.com") |>
  req_url_query(foo = c(1, 2)) |>
  req_dry_run()

#> GET /?foo=1&foo=2 HTTP/1.1
#> Host: example.com?foo=1&foo=2
#> User-Agent: httr2/0.2.3 r-curl/5.1.0 libcurl/8.3.0
#> Accept: */*
#> Accept-Encoding: deflate, gzip
```

I could see the other common form of multi-value parameters being handled via `I()`:

```r
# desired  behaviour
request("example.com") |>
  req_url_query(foo = I(c(1, 2))) |>
  req_dry_run()

#> GET /?foo=1,2 HTTP/1.1
#> Host: example.com?foo=1,2
#> User-Agent: httr2/0.2.3 r-curl/5.1.0 libcurl/8.3.0
#> Accept: */*
#> Accept-Encoding: deflate, gzip
```

hadley:
Fixes #350

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # httr2 (development version)
 
+* `req_url_query()` now allows you to set query parameters to a vector of 
+  values, in which case the name will be repeated, i.e. 
+  `req_url_query(req, a = 1)` will generate `?a=1&a=2` (#350).
+
 * The httr2 examples now only run on R 4.2 and later so that we can use
   the base pipe and lambda syntax (#345).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,8 +1,10 @@
 # httr2 (development version)
 
-* `req_url_query()` now allows you to set query parameters to a vector of 
-  values, in which case the name will be repeated, i.e. 
-  `req_url_query(req, a = 1)` will generate `?a=1&a=2` (#350).
+* `req_url_query()` gains a `.multi` parameter that controls what happens when
+  you supply multiple values in a vector. The default will continue to error 
+  but you can use `.multi = "comma"` to separate with commas, `"pipe"` to 
+  separate with `|`, and `"explode"` to generate one parameter for each 
+  value (e.g. `?a=1&a=2`) (#350).
 
 * The httr2 examples now only run on R 4.2 and later so that we can use
   the base pipe and lambda syntax (#345).
--- a/R/req-url.R
+++ b/R/req-url.R
@@ -32,7 +32,7 @@
 #' req |>
 #'   req_url("http://google.com")
 #'
-#' # Use multi to control what happens with vector parameters:
+#' # Use .multi to control what happens with vector parameters:
 #' req |> req_url_query(id = 100:105, .multi = "comma")
 #' req |> req_url_query(id = 100:105, .multi = "explode")
 #'
@@ -98,7 +98,7 @@ req_url_query <- function(.req,
     } else if (multi == "error") {
       cli::cli_abort(c(
         "All vector elements of {.code ...} must be length 1.",
-        i = "Use {.arg multi} to choose a strategy for handling."
+        i = "Use {.arg .multi} to choose a strategy for handling."
       ))
     }
   }
--- a/R/req-url.R
+++ b/R/req-url.R
@@ -9,10 +9,9 @@
 #' @inheritParams req_perform
 #' @param url New URL; completely replaces existing.
 #' @param ... For `req_url_query()`: <[`dynamic-dots`][rlang::dyn-dots]>
-#'   Name-value pairs that provide query parameters. Each value must be either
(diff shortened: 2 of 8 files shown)


NEWS.md
```diff
@@ -1,5 +1,9 @@
 # httr2 (development version)
 
+* `req_url_query()` now allows you to set query parameters to a vector of 
+  values, in which case the name will be repeated, i.e. 
+  `req_url_query(req, a = 1)` will generate `?a=1&a=2` (#350).
```
jonthegeek, line 5:
```suggestion
  `req_url_query(req, a = 1:2)` will generate `?a=1&a=2` (#350).
```

jonthegeek (approved):
I was JUST working around this about a week ago. Thanks!

hadley:
I'm much happier with this new approach.

R/req-url.R
```diff
@@ -32,6 +32,10 @@
 #' req |>
 #'   req_url("http://google.com")
 #'
+#' # Use multi to control what happens with vector parameters:
```
jonthegeek, line 35:
```suggestion
#' # Use .multi to control what happens with vector parameters:
```

jonthegeek (approved):
Just one typo. This will be super useful! Wonderful timing, since I just had to look up what the heck "explode" meant for an API definition last week!

Do you want to mention the function option in the news bullet?

jonthegeek:
BTW, this seems like the correct approach, but does it contradict https://design.tidyverse.org/independent-meaning.html ?

R/req-url.R
```diff
@@ -46,15 +50,79 @@ req_url <- function(req, url) {
...
+    } else if (multi == "pipe") {
+      dots[n > 1] <- lapply(dots[n > 1], format_query_param)
+      dots[n > 1] <- lapply(dots[n > 1], paste0, collapse = "|")
+      dots[n > 1] <- lapply(dots[n > 1], I)
+    } else if (multi == "explode") {
+      dots <- explode(dots)
+    } else if (multi == "error") {
+      cli::cli_abort(c(
+        "All vector elements of {.code ...} must be length 1.",
+        i = "Use {.arg multi} to choose a strategy for handling."
```
mgirlich, line 101: The info sentence feels unfinished (for handling what?). Maybe add something like "for handling multi element vectors".

mgirlich, line 101:
```suggestion
        i = "Use {.arg .multi} to choose a strategy for handling."
```

R/req-url.R
```diff
@@ -10,7 +10,7 @@
 #' @param url New URL; completely replaces existing.
 #' @param ... For `req_url_query()`: <[`dynamic-dots`][rlang::dyn-dots]>
 #'   Name-value pairs that provide query parameters. Each value must be either
-#'   a length-1 atomic vector (which is automatically escaped) or `NULL` (which
+#'   an atomic vector (which is automatically escaped) or `NULL` (which
```
mgirlich, line 13: That's actually not quite true that `NULL` is silently dropped. You can remove a query parameter with `NULL`:

``` r
request("https://example.com") |>
  req_url_query(foo = c(Sys.Date(), Sys.Date()), .multi = "explode") |>
  req_url_query(foo = NULL) |>
  req_dry_run()
#> GET / HTTP/1.1
#> Host: example.com
#> User-Agent: httr2/0.2.3.9000 r-curl/5.1.0 libcurl/8.1.2
#> Accept: */*
#> Accept-Encoding: deflate, gzip
```

<sup>Created on 2023-10-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

mgirlich (approved):

hadley:
@jonthegeek hmmm, maybe I'm splitting hairs, but I don't think this changes the meaning, it just changes how vectors are handled in the output.

jonthegeek:
@hadley I think you're correctly splitting hairs. It might be worth examining those hairs and mentioning them in that chapter, though 😊

PR #353, opened 2023-10-23
Implement new iteration strategy

### LINKED ISSUE #298: Store intermediate responses in `req_paginate()`
E.g. after an error or after user interrupt

### LINKED ISSUE #341: `iterate_next_request()` -> `resp_next_request()`

hadley:
A first pass. Fixes #341. Fixes #298.

To do (possibly in future PRs)

* [ ] What will `req_chunk()` (#307) look like?
* [x] Mechanism for signalling total number of pages (if known)
* [x] Do we need a way to return multiple pages (as in original)? — I don't think so? I think you could use `req_perform_parallel()` if you wanted.
* [ ] Should we return a new data structure containing both reqs and resps? If so, we should also do that in `req_perform_parallel()` (but not `multi_req_perform()`)

@mgirlich it's probably worth a glance if you have time, but it's likely to evolve quite a bit more as I work on it, so no need for a thorough review yet. It's a bit PR so I recommend reading `iterate.R` first, then `iterate-helpers.R` then `iterate-responses.R`


R/iterate-helpers.R
```diff
@@ -0,0 +1,82 @@
...
+#'   length(resp_body_json(resp)$data) == 0
+#' }
+#' resps <- req_perform_iteratively(
+#'   req,
+#'   iterate_with_offset("page_index", resp_complete = is_complete)
+#' )
+iterate_with_offset <- function(param_name,
+                                start = 1,
+                                offset = 1,
+                                resp_complete = NULL) {
```
hadley, line 37: `resp_complete()` is a different way to handle the maximum number of pages to request issue.

R/iterate-helpers.R
```diff
@@ -0,0 +1,115 @@
...
+  check_number_whole(offset, min = 1)
+  check_function2(resp_pages, args = "resp", allow_null = TRUE)
+  check_function2(resp_complete, args = "resp", allow_null = TRUE)
+  resp_complete <- resp_complete %||% function(resp) FALSE
+
+  known_total <- FALSE
+  i <- start # assume already fetched
+
+  function(resp, req) {
+    if (!is.null(resp_pages) && !known_total) {
```
hadley, line 71: I like how this implementation using signals turned out because I think most folks will end up using `req_perform_iteratively()` with a canned helper and this makes the implementation much simpler.

R/iterate.R
```diff
@@ -0,0 +1,118 @@
...
+      withCallingHandlers(
+        {
+          req <- next_req(resp = resp, req = req)
+        },
+        httr2_total_pages = function(cnd) {
+          # Allow next_req() to shrink the number of pages remaining
+          # Most important in max_req = Inf case
+          if (cnd$n < max_reqs) {
+            max_reqs <<- cnd$n
+            progress$update(total = max_reqs)
```
mgirlich, line 86: FYI I think you need to specify `inc = 0`, otherwise the progress bar is also incremented. At least I remember encountering this issue somewhere and it took me some time to figure out why the progress bar failed in some cases...

R/utils.R
```diff
@@ -271,5 +274,10 @@ create_progress_bar <- function(total,
   args$total <- total
   args$.envir <- env
 
-  exec(cli::cli_progress_bar, !!!args)
+  id <- exec(cli::cli_progress_bar, !!!args)
+
+  list(
+    update = function(...) cli::cli_progress_update(..., id = id),
+    done = function() cli::cli_progress_done(id = id)
```
mgirlich, line 281: That's a good idea!

mgirlich:
I quickly skimmed over it and it looks quite nice. Let me know when this is ready for a proper review.

R/iterate.R
hadley: Good catch, thanks.

hadley:
I think it's probably good for a proper review now. The part I'm least certain about is the return value — my gut feeling is that it should return a list of resps and a list of reqs, so that if there is an error you can track it down more easily. But I'm struggling with what to call that object, and it should go in a separate PR anyway.

R/iterate.R
```diff
@@ -0,0 +1,118 @@
...
+#' `req_perform_iteratively()` iteratively generates and performs requests,
+#' using a callback function, `next_req`, to define the next request based on
+#' the current request and response. You will probably want to it pair with an
+#' [iteration helper][iterate_with_offset] and use a
+#' [multi-response handler][resps_combine] to process the result.
+#'
+#' @inheritParams req_perform
+#' @param next_req A function that takes the previous response (`resp`) and
+#'   request (`req`) and returns a [request] for the next page or `NULL` if
+#'   the iteration should terminate.
```
mgirlich, line 15: This should mention how to signal the number of pages and we might want to provide a helper for that.

R/iterate.R
```diff
@@ -0,0 +1,118 @@
...
+      progress$update()
+
+      withCallingHandlers(
+        {
+          req <- next_req(resp = resp, req = req)
+        },
+        httr2_total_pages = function(cnd) {
+          # Allow next_req() to shrink the number of pages remaining
+          # Most important in max_req = Inf case
+          if (cnd$n < max_reqs) {
```
mgirlich, line 84: Check `cnd$n`. I had a typo in the field with the number of pages and signalled an empty numeric resulting in a bad error message.

R/iterate.R
mgirlich: Also we might provide an example (including the technique with `known_total`) or point the user to where to look this up.

mgirlich:
I used the new interface in an internal package (for an absolutely terrible API with a very inconsistent interface. So, the perfect use case actually). Some things that might be difficult:

* the token is sometimes in a nested field. This forced me to used `purrr::pluck()`
* the token must be send in the body 🙄 which is quite ugly. I'm not sure this is often the case but it might be nice to add a helper to modify the body
* I copied the part how to signal the number of pages. We probably want to add a helper for this.

```r
api_paginate_token <- function(req,
                               field,
                               spec,
                               page_size,
                               n_results_field,
                               ...,
                               req_token_field = "nextToken",
                               resp_token_field = "nextToken",
                               pb_name = "Download",
                               n = Inf) {
  check_dots_empty()

  parse_resp <- function(resp) {
    parsed <- httr2::resp_body_json(resp, check_type = FALSE, bigint_as_char = TRUE)
    tibblify(parsed[[field]], spec)
  }

  known_total <- FALSE
  next_req <- function(resp, req) {
    parsed <- httr2::resp_body_json(resp, check_type = FALSE, bigint_as_char = TRUE)
    next_token <- purrr::pluck(parsed, !!!resp_token_field)

    if (is.null(next_token)) {
      return(NULL)
    }

    if (!known_total && !is.null(n_results_field)) {
      n_results <- purrr::pluck(parsed, !!!n_results_field)
      check_number_whole(n_results)
      known_total <<- TRUE
      signal("", class = "httr2_total_pages", n = ceiling(n_results / page_size))
    }

    data <- req$body$data
    data <- purrr::assign_in(data, req_token_field, value = u(next_token))
    req$body$data <- data
    req
  }

  responses <- httr2::req_perform_iteratively(
    req = req,
    next_req = next_req,
    max_reqs = ceiling(n / page_size),
    progress = pb_name
  )

  httr2::resps_combine(responses, parse_resp)
}
```

mgirlich:
Oh and of course I couldn't use the nice helpers because of the pagination patterns used by the API...

R/iterate-helpers.R
```diff
@@ -0,0 +1,115 @@
...
+#'
+#' @description
+#' These functions are suitable for use with the `next_req` argument to
+#' [req_perform_iteratively()]. Each implements iteration for a common
+#' pagination pattern:
+#'
+#' * `iterate_with_offset()` increments a query parameter, e.g. `?page=1`,
+#'   `?page=2`, or `?offset=1`, `offset=21`.
+#' * `iterate_with_cursor()` updates a query parameter with the value of a
+#'    cursor found somewhere in the response.
```
mgirlich, line 11: Restricting `iterate_with_offset()` and `iterate_with_cursor()` to a query parameter makes the implementation much simpler. I hope this covers most APIs but it definitely doesn't cover the main one I have to use...

R/iterate-helpers.R
```diff
@@ -0,0 +1,115 @@
+#' Iteration helpers
+#'
+#' @description
+#' These functions are suitable for use with the `next_req` argument to
```
mgirlich, line 4:
```suggestion
#' These functions are intended for use with the `next_req` argument to
```

R/iterate-helpers.R
```diff
@@ -0,0 +1,115 @@
...
+#' [req_perform_iteratively()]. Each implements iteration for a common
+#' pagination pattern:
+#'
+#' * `iterate_with_offset()` increments a query parameter, e.g. `?page=1`,
+#'   `?page=2`, or `?offset=1`, `offset=21`.
+#' * `iterate_with_cursor()` updates a query parameter with the value of a
+#'    cursor found somewhere in the response.
+#'  * `iterate_with_link_url()` follows the url found the `Link` header.
```
mgirlich, line 12:
```suggestion
#'  * `iterate_with_link_url()` follows the url found in the `Link` header.
```

R/iterate-helpers.R
```diff
@@ -0,0 +1,115 @@
...
+#'    See `resp_link_url()` for more details.
+#'
+#' @param param_name Name of query parameter.
+#' @param start Starting value.
+#' @param offset Offset for each page.
```
mgirlich, line 17: Maybe explain that the default 1 works if paginating by page index and otherwise this should be the page size.

R/iterate-helpers.R
```diff
@@ -0,0 +1,115 @@
...
+#'   be called once.
+#' @export
+#' @examples
+#' req <- request(example_url()) |>
+#'   req_url_path("/iris") |>
+#'   req_throttle(10) |>
+#'   req_url_query(limit = 50)
+#'
+#' # If you don't know the total number of pages in advance, you can
+#' # provde a `resp_complete()` callback
```
mgirlich, line 31:
```suggestion
#' # provide a `resp_complete()` callback
```

R/iterate-helpers.R
```diff
@@ -0,0 +1,115 @@
...
+#' # can easily calculate it), you can use the `resp_pages()` callback which
+#' # will generate a better progress bar.
+#'
+#' resps <- req_perform_iteratively(
+#'   req %>% req_url_query(limit = 1),
+#'   next_req = iterate_with_offset(
+#'     "page_index",
+#'     resp_pages = function(resp) resp_body_json(resp)$pages
+#'   ),
+#'   max_reqs = Inf)
```
mgirlich, line 52: `)` should be on its own line

R/iterate-helpers.R
```diff
@@ -0,0 +1,115 @@
...
+    }
+  }
+}
+
+#' @rdname iterate_with_offset
+#' @export
+#' @param resp_param_value A callback function that takes a response (`resp`)
+#'   and returns the next cursor value. Return `NULL` if there are no further
+#'   pages.
+iterate_with_cursor <- function(param_name, resp_param_value) {
```
mgirlich, line 91: Is there a reason to not include a `resp_pages` argument?

R/iterate-helpers.R
```diff
@@ -0,0 +1,115 @@
...
+    if (!is.null(value)) {
+      req %>% req_url_query(!!param_name := value)
+    }
+  }
+}
+
+#' @rdname iterate_with_offset
+#' @export
+#' @param rel The "link relation type" to use to retrieve the next page.
+iterate_with_link_url <- function(rel = "next") {
```
mgirlich, line 106: Is there a reason to not include a `resp_pages` argument?

R/iterate-helpers.R
```diff
@@ -0,0 +1,115 @@
...
+#' @param resp_param_value A callback function that takes a response (`resp`)
+#'   and returns the next cursor value. Return `NULL` if there are no further
+#'   pages.
+iterate_with_cursor <- function(param_name, resp_param_value) {
+  check_string(param_name)
+  check_function2(resp_param_value, args = "resp")
+
+  function(resp, req) {
+    value <- resp_param_value(resp)
+    if (!is.null(value)) {
```
mgirlich, line 97: Not really necessary but you could check `value` (i.e. it is a scalar atomic)

R/iterate-responses.R
```diff
@@ -0,0 +1,34 @@
+#' Tools for working with lists of responses
+#'
+#' * `resps_combine()` combines the data from each responses into a single
+#'   object.
+#' * `resps_response()` returns all successful responses.
```
mgirlich, line 5: Maybe `resps_success()` to better match `resps_error()`?

R/iterate-responses.R
```diff
@@ -0,0 +1,34 @@
+#' Tools for working with lists of responses
+#'
+#' * `resps_combine()` combines the data from each responses into a single
```
mgirlich, line 3:
```suggestion
#' * `resps_combine()` combines the data from each response into a single
```

R/iterate-responses.R
```diff
@@ -0,0 +1,34 @@
...
+#'   object.
```
mgirlich, line 4:
```suggestion
#'   vector.
```

R/iterate-responses.R
```diff
@@ -0,0 +1,34 @@
...
+#' @param resp_data A function that takes a response (`resp`) and
+#'   returns its data as a vector or data frame.
+#' @param resp_data A function with one argument `resp` that parses the
+#'   response and returns a list with the field `data` and other fields needed
+#'   to create the request for the next page.
+resps_combine <- function(resps, resp_data) {
+  check_installed("vctrs")
+
+  check_function2(resp_data, "body")
+  vctrs::vec_c(!!!lapply(resps, resp_data))
```
mgirlich, line 19:
```suggestion
  vctrs::list_unchop(lapply(resps, resp_data))
```

R/iterate-responses.R
```diff
@@ -0,0 +1,34 @@
...
+#' * `resps_response()` returns all successful responses.
+#' * `resps_error()` returns all errors.
+#'
+#' @export
+#' @param resps A list of responses (possibly including errors).
+#' @param resp_data A function that takes a response (`resp`) and
+#'   returns its data as a vector or data frame.
+#' @param resp_data A function with one argument `resp` that parses the
+#'   response and returns a list with the field `data` and other fields needed
+#'   to create the request for the next page.
```
mgirlich, line 14:
```suggestion
```

R/iterate-responses.R
```diff
@@ -0,0 +1,34 @@
...
+#'
+#' * `resps_combine()` combines the data from each responses into a single
+#'   object.
+#' * `resps_response()` returns all successful responses.
+#' * `resps_error()` returns all errors.
+#'
+#' @export
+#' @param resps A list of responses (possibly including errors).
+#' @param resp_data A function that takes a response (`resp`) and
+#'   returns its data as a vector or data frame.
```
mgirlich, line 11:
```suggestion
#'   returns it's data as a vector or data frame.
```

R/iterate-responses.R
```diff
@@ -0,0 +1,34 @@
...
+#' @param resp_data A function with one argument `resp` that parses the
+#'   response and returns a list with the field `data` and other fields needed
+#'   to create the request for the next page.
+resps_combine <- function(resps, resp_data) {
+  check_installed("vctrs")
+
+  check_function2(resp_data, "body")
```
mgirlich, line 18: I'm pretty sure the argument (also as documented above) is `resp`, not `body`

```suggestion
  check_function2(resp_data, "resp")
```

R/iterate-responses.R
```diff
@@ -0,0 +1,34 @@
...
+#' * `resps_error()` returns all errors.
+#'
+#' @export
+#' @param resps A list of responses (possibly including errors).
+#' @param resp_data A function that takes a response (`resp`) and
+#'   returns its data as a vector or data frame.
+#' @param resp_data A function with one argument `resp` that parses the
+#'   response and returns a list with the field `data` and other fields needed
+#'   to create the request for the next page.
+resps_combine <- function(resps, resp_data) {
```
mgirlich, line 15: CRAN might be more happy if you explicitly document the return value with `@return`

R/iterate.R
```diff
@@ -0,0 +1,118 @@
...
+#' @inheritParams req_perform
+#' @param next_req A function that takes the previous response (`resp`) and
+#'   request (`req`) and returns a [request] for the next page or `NULL` if
+#'   the iteration should terminate.
+#' @param max_reqs The maximum number of requests to perform. Use `Inf` to
+#'   perform all requests until `next_req()` returns `NULL`.
+#' @param progress Display a progress bar? Use `TRUE` to turn on a basic
+#'   progress bar, use a string to give it a name, or see [progress_bars] to
+#'   customise it in other ways.
+#' @param path Optionally, path to save body of request. This should be
```
mgirlich, line 21:
```suggestion
#' @param path Optionally, path to save the body of request. This should be
```

(the same applies for `req_perform()`)

R/iterate.R
```diff
@@ -0,0 +1,118 @@
...
+#' })
+req_perform_iteratively <- function(req,
+                                    next_req,
+                                    path = NULL,
+                                    max_reqs = 20,
+                                    progress = TRUE) {
+  check_request(req)
+  check_function2(next_req, args = c("resp", "req"))
+  check_number_whole(max_reqs, allow_infinite = TRUE, min = 1)
+  check_string(path, allow_empty = FALSE, allow_null = TRUE)
```
mgirlich, line 53: Maybe check that the path actually includes `{i}`?

R/test.R
```diff
@@ -59,12 +33,12 @@ example_url <- function() {
...
-    end <- min(start + page_size - 1L, n)
-    ids <- seq(start, end)
-    data <- vctrs::vec_slice(datasets::iris, ids)
+    end <- start + page_size - 1L
+    ids <- seq2(start, end)
+    data <- vctrs::vec_slice(datasets::iris, intersect(ids, seq_len(n)))
 
     res$set_status(200L)$send_json(
-      object = list(data = data, count = n),
+      object = list(data = data, count = n, pages = n / page_size),
```
mgirlich, line 41:
```suggestion
      object = list(data = data, count = n, pages = ceiling(n / page_size)),
```

mgirlich, line 41:
```suggestion
      object = list(data = data, count = n, pages = ceiling(n / page_size)),
```

mgirlich (approved):

mgirlich:
Overall this looks good to me. Switching to the new interface was relatively quick (as written above there are some annoying things but I guess this is mainly due to the terrible API I have to use).

mgirlich:
> The part I'm least certain about is the return value — my gut feeling is that it should return a list of resps and a list of reqs, so that if there is an error you can track it down more easily. But I'm struggling with what to call that object, and it should go in a separate PR anyway.

Maybe add the request to each response as an attribute. Or add a helper `last_iterative_requests()`.

R/iterate-responses.R
hadley: Although I often get this wrong I'm pretty sure I'm right here 😄

hadley: But maybe the wording is confusing.

mgirlich: Ah, so true, you're right, it's "its" 😊

R/iterate-helpers.R
hadley: I'll just rely on `req_url_query()` for this, I think.

R/iterate.R
hadley: I'll do that when I make a helper.

R/iterate-helpers.R
hadley: I think it's a reasonable trade-off. My plan is to eventually have vignette that describes how to handle more complex situations.

R/iterate-helpers.R
hadley: I didn't include a `resp_pages()` for either of these two helpers because I think it's relatively rare that the total number of pages is available with these interfaces. But I guess there's no harm in adding it.

R/iterate-responses.R
hadley: I was thinking of these more about as the names of the types you get back, rather success vs failures. Anyway, I'll iterate on these as part of #357.

R/iterate.R
hadley: Tracking in #358 — I think I'm likely to move the "do this once" check to `req_perform_iteratively()`.

R/iterate.R
hadley: Thats a bit fiddly for me for now, particularly since I'm not sure if I'm super committed to this API. We can work on it in the future.

PR #360, opened 2023-10-26
Improve iteration documentation

### LINKED ISSUE #358: Explain iteration in more detail
Including how to signal the total number of pages. Export a helper to make the signal.

hadley:
Includes two new functions: `req_body_json_modify()` and `signal_total_pages()`.

Fixes #358


R/iterate.R
```diff
@@ -9,6 +9,62 @@
...
+#' like this:
+#'
+#' ```R
+#' next_req <- function(resp, req) {
+#'   body <- resp_body_json(resp)
+#'   cursor <- body$next_cursor
+#'   if (is.null(cursor))
+#'     return(NULL)
+#'
+#'   signal_total_pages(body$pages)
```
hadley, line 63: I don't think it's really that important to only signal this once. We could probably remove it from  `iterate_with_offset()` but I feel like it's worth being a bit more cautious in a helper, but it's generally hard to imagine that this is going to be expensive to compute.

R/req-body.R
```diff
@@ -106,14 +107,30 @@ req_body_json <- function(req, data,
 
 #' @export
 #' @rdname req_body
-#' @param ... <[`dynamic-dots`][rlang::dyn-dots]> Name-data pairs used send
-#'   data in the body. For `req_body_form()`, the values must be strings (or
-#'   things easily coerced to string); for `req_body_multipart()` the values
-#'   must be strings or objects produced by
-#'   [curl::form_file()]/[curl::form_data()].
+req_body_json_modify <- function(req, ...) {
```
hadley, line 110: To avoid accessing the internals like in your example.

mgirlich: Is there any way I can modify a nested value?

E.g. how do I modify the value of `b` in `list(a = list(b = 1))`? If I can't do that then I need to somehow extract `a`, then modify `b` and then assign `a`.
There is a reason I used `assign_in()` in my example... This devilish API hides the cursor token in `c("pageCriteria", "identifier", "token")` and this feels like an example of over-engineering or pure evil 😞

tests/testthat/test-req-body.R
```diff
@@ -81,6 +81,14 @@ test_that("non-json type errors", {
   )
 })
 
+test_that("can modify json data", {
+  req <- request_test() %>%
+    req_body_json(data = list(a = 1, b = 2, d = 4)) %>%
+    req_body_json_modify(a = 10, b = NULL, c = 3)
+
+   expect_equal(req$body$data, list(d = 4, a = 10, c = 3))
```
mgirlich, line 89: Not really relevant but it might be nice if `req_body_json_modify()` would keep the order of the list, i.e.

```r
   expect_equal(req$body$data, list(a = 10, c = 3, d = 4))
```

mgirlich (approved):

R/req-body.R
hadley: I think `modifyList()` handles that case, but not `modify_list()`, which isn't really right for modifying json anyway, since it's designed to allow you to set multiple values with the same name. I'll dig into this a bit more.

PR #361, opened 2023-10-26
First stab at req_perform_sequential()

### LINKED ISSUE #303: Add helper to perform a request in chunks

### LINKED ISSUE #310: Add `error` argument to `req_perform_iteratively()`
In case a request fails it might still be interesting to continue (if possible) with the pagination or return the intermediate results.

hadley:
Fixes #303. Fixes #310

@mgirlich this is just a quick proof of concept. If you think it looks good, I'll finish off the docs and tests. I did try allowing a `NULL` `req` in `req_perform_iteratively()`, but it ended up feeling very artificial since the `next_req()` callback would then ignore both its arguments. 

If we have `req_perform_parallel()`, `req_perform_sequential()`, then I think the matching name would be `req_perform_iterative()`, not iteratively.


R/sequential.R
```diff
@@ -0,0 +1,48 @@
...
+  resps <- rep_along(reqs, list())
+
+  tryCatch({
+    for (i in seq_along(reqs)) {
+      check_request(reqs[[i]], arg = glue::glue("req[[{i}]]"))
+      resps[[i]] <- req_perform(reqs[[i]], path = paths[[i]])
+      progress$update()
+    }
+  }, interrupt = function(cnd) {
+    cli::cli_alert_warning(
```
mgirlich, line 41: When is the interrupt actually caught? I always seem to get

```
Error in `req_perform()` at httr2/R/sequential.R:37:6:  1s
! Failed to perform HTTP request.
Caused by error in `curl::curl_fetch_memory()`:
! Operation was aborted by an application callback
```

when I interrupt the example with the escape key.

mgirlich:
Looks good to me. We need to think about error handling. It should probably be the same as in `req_perform_parallel()` though I'm not sure we want to stick with the name `cancel_on_error` and the boolean logic there.

hadley:
Yeah, me have a go at #310. I think an enum is the way forward. I've written up https://design.tidyverse.org/boolean-strategies.html to hopefully help me avoid this problem again in the future.

R/sequential.R
hadley: Oh shoot, I somehow never hit that when I tested this before, but now it's all I see 😢

hadley: And I already filed an issue on this in curl: https://github.com/jeroen/curl/issues/271. I guess I'll leave this code here since it will eventually be useful but there's not much we can do until then.

R/sequential.R
```diff
@@ -0,0 +1,94 @@
+#' Perform multiple requests in sequence
+#'
+#' Given a list of requests, this function performs each in turn, returning
+#' a list of responses. It's slower than [req_perform_parallel()] but
+#' has fewer limitations.
+#'
+#' @inheritParams req_perform_parallel
+#' @inheritParams req_perform_iteratively
+#' @param on_error What should happen if one of the requests throws an
```
hadley, line 9: @mgirlich let me know what you think of this error handling, and if it looks good, I'll port to `req_perform_parallel()`.

hadley: And `req_perform_iterative()`, although that won't have the `"continue"` option.

R/sequential.R
```diff
@@ -0,0 +1,94 @@
...
+#'   HTTP error?
+#'
+#'   * `stop`, the default: stop iterating and throw an error
+#'   * `return`: stop iterating and return all the successful responses so far.
+#'   * `ignore`: continue iterating, recording errors in the result.
```
mgirlich, line 14: `on_error = "ignore"` doesn't read as nicely as with `stop` and `return`. I think `on_error = "continue"` would be a bit more clear (and you already mentioned the "continue" option above anyways).

R/sequential.R
```diff
@@ -0,0 +1,94 @@
...
+#' @export
+#' @return
+#' A list, the same length as `reqs`.
+#'
+#' If `error_on` is `"return"` and it errors on the ith request, the ith
```
mgirlich, line 19:
```suggestion
#' If `on_error` is `"return"` and it errors on the ith request, the ith
```

R/sequential.R
```diff
@@ -0,0 +1,94 @@
...
+#' element of the result will be an error object, and the remaining elements
+#' will be `NULL`.
+#'
+#' If `error_on` is `"ignore"`, it will be a mix of requests and errors.
```
mgirlich, line 23:
```suggestion
#' If `on_error` is `"ignore"`, it will be a mix of requests and errors.
```

R/sequential.R
```diff
@@ -0,0 +1,94 @@
...
+#' }
+req_perform_sequential <- function(reqs,
+                                   paths = NULL,
+                                   on_error = c("stop", "return", "ignore"),
+                                   progress = TRUE) {
+  if (!is_bare_list(reqs)) {
+    stop_input_type(reqs, "a list")
+  }
+  if (!is.null(paths)) {
+    if (length(reqs) != length(paths)) {
```
mgirlich, line 54: Maybe also check the type with `check_character()`? I also opened an issue for `req_perform()` #364.

mgirlich (approved):

R/sequential.R
hadley: Agreed

PR #374, opened 2023-11-05
Render port in getting started vignette

jonthegeek:
A typo `R` for `r` causes the port to not render as intended.


hadley:
Thanks!

PR #375, opened 2023-11-06
Use "Uploading" for type "up"

mgirlich:
For `type = "up"` it now says "Uploading" instead of "Downloading".
@hadley do you want some tests for `make_progress()`? There aren't any yet at all.


hadley:
Yeah, some tests would be good. I wonder if there's a way to just test the final display.

PR #383, opened 2023-11-10
Update phantom httr references from httr to httr2.

jonthegeek:
The libcurl version updated when I reknitted README.Rmd. It made me notice that you'll want to be sure to reknit the README after the version gets bumped to pick up the proper version in the User-Agent, which I don't think is a normal part of the checklist.


hadley:
Thanks! I don't care too much about ensuring that UA string is precisely correct in the readme.

jonthegeek:
> Thanks! I don't care too much about ensuring that UA string is precisely correct in the readme.

Yeah, it's more just to avoid questions from confused users 🙃

PR #394, opened 2023-11-17
Use _ instead of . with base pipe

MSHelm:
Hey everyone,
thanks for the great package! While reading through the docs I noticed that the wrapping API article uses invalid base pipe syntax, fixed here.

I also noticed that the articles do not check for availability of the base pipe for evaluation of the code chunks, like you do in the vignette. If this should be done I would be happy to include this here as well :)


hadley:
Thanks! No need to check here because articles aren't run on older versions of R, unlink the vignettes.

PR #399, opened 2023-11-18
Move oauth vignette out of articles.

### LINKED ISSUE #398: oauth vignette is an article
Several help docs mention `vignette("oauth")`, but that vignette is only available on the website. Trying to load it throws a warning, and doesn't do anything:

``` r
vignette("oauth", package = "httr2")
#> Warning: vignette 'oauth' not found
```

<sup>Created on 2023-11-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

I'm not certain yet whether it would be as simple as moving the vignette out of articles, or if that vignette can't build and the help docs instead need to be tweaked (or if I have something configured strangely to cause the error in the first place).

jonthegeek:
Closes #398.

I'm not sure if there's something not-CRAN-safe that prevents this from being here, but it seems to work locally at least.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,9 @@
 * `req_template()` now works when you have a bare `:` in a template that
   uses "uri" style (#389).
 
+* `vignette("oauth")` is now available as a package vignette, not just on the 
+  package website (@jonthegeek, #398).
+
 # httr2 1.0.0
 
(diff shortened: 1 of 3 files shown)


hadley:
I don't want this as a vignette as HTTP using vignettes are high risk on CRAN.

PR #410, opened 2023-12-12
use rstudioapi instead of readline if calling oauth_flow_auth_code() from RStudio IDE

### LINKED ISSUE #406: Enable user response when `oauth_flow_auth_code()` is called indirectly
While using `oauth_token_cached()` with `flow = oauth_flow_auth_code` and `cache_disk = TRUE` (as added in #349):

1. ~99% of calls successfully leverage the cache (or reauthenticate in the background)
2.  ~ 1% of calls trigger a new prompt

Each are the expected behavior, which has been immensely helpful! 

But, a nuance we've since realized is that if (2) happens when called **indirectly**: 

1. _sequential R chunks run in .Rmd_: doesn't always wait for the user to enter a code before trying to run the subsequent code. FWIW I've noticed the same in `gargle` per https://github.com/r-lib/gargle/blob/main/R/oauth-init.R#L133
2. _knitted .Rmd_: won't run, since it's non-interactive

(1) is a side effect of the [expected and well-documented behavior of `readline()`](https://stackoverflow.com/questions/27112370/make-readline-wait-for-input-in-r) that it doesn't necessarily wait for use input at all costs. 

One option could be using an `rstudioapi` alternate in `oauth_flow_auth_code_read()`, but that's not as generalizable:
```
oauth_flow_auth_code_read <- function(state) {
  prompt <- "Enter authorization code or URL: "
  code <- rstudioapi::askForPassword(prompt)
  code <- trimws(code)
...
}
```

(2) Could likely be offset by changing https://github.com/r-lib/httr2/blob/main/R/oauth-flow-auth-code.R#L145, akin to #171

Here's one way to reproduce (1), to start:
1. Creating a new .Rmd file with two chunks:

```{r chunk_1}
# googlesheets4::gs4_auth()
# or
httr2:::oauth_flow_auth_code_read("state")
```

```{r chunk_2}
summary(cars)
```

2. Run All (not knit) 
image

Sends to the R Console:
```
> # Chunk 1
> # googlesheets4::gs4_auth()
> # or
> httr2:::oauth_flow_auth_code_read("state")
Enter authorization code or URL: 
Enter state parameter: # Chunk 2
Error in `httr2:::oauth_flow_auth_code_read()`:
! Authentication failure: state does not match
```

fh-mthomson:
Closes #406 

# Test plan

Running `oauth_flow_auth_code_read("state")` successfully prompts twice (initial prompt, then state) in either case described in parent issue:
1. sequential R chunks run in .Rmd 
2. knitting .Rmd

# Implementation considerations
* Initially implemented a version of using _only_ `askpass`, given broader generalizability outside of RStudio IDE; however, this doesn't scale as well off the shelf to the non-interactive use case (e.g., knitting in RStudio) - per `?askpass::askpass`:
```
By default askpass() returns NULL in non-interactive sessions. (These include knitr runs and testthat tests.) If you want to force a password prompt in non-interactive sessions, set the rlib_interactive option to TRUE:
options(rlib_interactive = TRUE)
```
* Opted instead for an initial implementation of `rstudioapi` > `askpass`, but this may be too layered - open to any feedback!
* Also considered [getPass](https://github.com/wrathematics/getPass) but it seems less mature / robust.

Note: leaving [this instance of readline()](https://github.com/r-lib/httr2/blob/10f3d878da94f2597c02c29846d446f3fbd635c8/R/oauth-flow-device.R#L76) since it's not really used to solicit a character string from the user and is likely better handled by another prompt function, anyway.

### DIFF
--- a/R/oauth-flow-auth-code.R
+++ b/R/oauth-flow-auth-code.R
@@ -106,6 +106,7 @@ req_oauth_auth_code <- function(req,
                                 host_name = deprecated(),
                                 host_ip = deprecated(),
                                 port = deprecated()) {
+
   redirect <- normalize_redirect_uri(
     redirect_uri = redirect_uri,
     host_name = host_name,
@@ -138,7 +139,9 @@ oauth_flow_auth_code <- function(client,
                                  redirect_uri = oauth_redirect_uri(),
                                  host_name = deprecated(),
                                  host_ip = deprecated(),
-                                 port = deprecated()) {
+                                 port = deprecated()
+ ) {
+
   oauth_flow_check("authorization code", client, interactive = TRUE)
 
   redirect <- normalize_redirect_uri(
@@ -201,6 +204,7 @@ normalize_redirect_uri <- function(redirect_uri,
                                    host_ip = deprecated(),
                                    port = deprecated(),
                                    error_call = caller_env()) {
+
   parsed <- url_parse(redirect_uri)
 
   if (lifecycle::is_present(host_name)) {
@@ -246,6 +250,7 @@ normalize_redirect_uri <- function(redirect_uri,
     localhost = localhost,
     can_fetch_code = can_fetch_oauth_code(redirect_uri)
   )
+
 }
 
 
@@ -398,27 +403,10 @@ oauth_flow_auth_code_pkce <- function() {
   )
 }
 
-# Try to determine whether we can redirect the user's browser to a server on
-# localhost, which isn't possible if we are running on a hosted platform.
-#
-# Currently this detects RStudio Server, Posit Workbench, and Google Colab. It
-# is based on the strategy pioneered by the {gargle} package.
-is_hosted_session <- function() {
-  if (nzchar(Sys.getenv("COLAB_RELEASE_TAG"))) {
-    return(TRUE)
-  }
-  # If RStudio Server or Posit Workbench is running locally (which is possible,
-  # though unusual), it's not acting as a hosted environment.
-  Sys.getenv("RSTUDIO_PROGRAM_MODE") == "server" &&
-    !grepl("localhost", Sys.getenv("RSTUDIO_HTTP_REFERER"), fixed = TRUE)
-}
 
-is_rstudio_session <- function() {
-  !is.na(Sys.getenv("RSTUDIO_PROGRAM_MODE", unset = NA))
-}
 
 oauth_flow_auth_code_read <- function(state) {
-  code <- ask_user_prompt("Enter authorization code or URL: ")
+  code <- prompt_user("Enter authorization code or URL: ")
 
   if (is_string_url(code)) {
     # minimal setup where user copy & pastes a URL
@@ -438,7 +426,7 @@ oauth_flow_auth_code_read <- function(state) {
     # Full manual approach, where the code and state are entered
     # independently.
 
-    new_state <- ask_user_prompt("Enter state parameter: ")
+    new_state <- prompt_user("Enter state parameter: ")
   }
 
   if (!identical(state, new_state)) {
--- a/R/oauth-flow-auth-code.R
+++ b/R/oauth-flow-auth-code.R
@@ -419,7 +419,8 @@ is_hosted_session <- function() {
 }
 
 oauth_flow_auth_code_read <- function(state) {
-  code <- trimws(readline("Enter authorization code or URL: "))
+  check_installed("askpass")
+  code <- askpass::askpass("Enter authorization code or URL")
 
   if (is_string_url(code)) {
     # minimal setup where user copy & pastes a URL
@@ -439,7 +440,7 @@ oauth_flow_auth_code_read <- function(state) {
     # Full manual approach, where the code and state are entered
     # independently.
 
-    new_state <- trimws(readline("Enter state parameter: "))
+    new_state <- askpass::askpass("Enter state parameter")
   }
 
   if (!identical(state, new_state)) {
@@ -486,6 +487,3 @@ oauth_flow_auth_code_fetch <- function(state) {
   body <- resp_body_json(resp)
   body$code
 }
-
-# Make base::readline() mockable
-readline <- NULL
--- a/R/oauth-flow-auth-code.R
+++ b/R/oauth-flow-auth-code.R
@@ -140,7 +140,7 @@ oauth_flow_auth_code <- function(client,
                                  host_name = deprecated(),
                                  host_ip = deprecated(),
                                  port = deprecated()
- ) {
+) {
 
   oauth_flow_check("authorization code", client, interactive = TRUE)
 
--- a/R/oauth-flow-auth-code.R
+++ b/R/oauth-flow-auth-code.R
@@ -106,7 +106,6 @@ req_oauth_auth_code <- function(req,
                                 host_name = deprecated(),
                                 host_ip = deprecated(),
                                 port = deprecated()) {
-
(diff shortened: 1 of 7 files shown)


jennybc:
In case it's relevant, gargle uses `readline()` for very specific reasons, which you can read about here:

https://github.com/r-lib/gargle/pull/242

The short version is to support Jupyter notebooks and, in particular, Google Colab.

vignettes/articles/wrapping-apis.Rmd
```diff
@@ -501,13 +501,13 @@ You can make this approach a little more user friendly by providing a helper tha
...
   if (is.null(key)) {
-    key <- askpass::askpass("Please enter your API key")
+    key <- rstudioapi::askForPassword("Please enter your API key")
   }
   Sys.setenv("NYTIMES_KEY" = key)
 }
 ```
 
-Using askpass (or similar) here is good practice since you don't want to encourage the user to type their secret key into the console, as mentioned above.
+Using `rstudioapi` (or similar) here is good practice since you don't want to encourage the user to type their secret key into the console, as mentioned above.
```
fh-mthomson, line 510: Still valid advice, but then we partially erode the security benefits by immediately calling `Sys.setenv()` 🤔 
https://github.com/rstudio/rsconnect/pull/1027#issuecomment-1820901085

In any case, if we decide to use `rstudioapi` over `askpass`, feels prudent to remove usage of `askpass` here (thereby dropping it from `Suggests` as well).

fh-mthomson:
> In case it's relevant, gargle uses `readline()` for very specific reasons, which you can read about here:
> 
> [r-lib/gargle#242](https://github.com/r-lib/gargle/pull/242)
> 
> The short version is to support Jupyter notebooks and, in particular, Google Colab.

Super relevant! I simplified a bit to only use `rstudioapi` if running in RStudio IDE; otherways, stick with `readline()`. Lmk if this strikes a better balance.

I do think this PR is a meaningful improvement; it would prevent folks on our end from missing an Okta re-auth prompt several times a day.

vignettes/articles/wrapping-apis.Rmd
hadley: Calling `Sys.setenv()` this was is fine, because it won't end up in the `.Rhistory` (because it's both inside a function and even if called at the top level it'd only record `key`, not the actual value).

R/oauth-flow-password.R
```diff
@@ -69,9 +69,20 @@ oauth_flow_password <- function(client,
...
-    password <- askpass::askpass()
+    password <- ask_user_prompt()
   }
   check_string(password, call = call)
   password
 }
+
+ask_user_prompt <- function(prompt = "Please enter your password: ") {
+  if (is_rstudio_session()) {
+    check_installed("rstudioapi")
```
hadley, line 80: Might be nice to use the `reason` argument to `check_installed()` to give the user some context

R/oauth-flow-auth-code.R
```diff
@@ -418,8 +413,12 @@ is_hosted_session <- function() {
     !grepl("localhost", Sys.getenv("RSTUDIO_HTTP_REFERER"), fixed = TRUE)
 }
 
+is_rstudio_session <- function() {
```
hadley, line 416: Should live by `ask_user_prompt()`

R/oauth-flow-auth-code.R
```diff
@@ -106,7 +106,6 @@ req_oauth_auth_code <- function(req,
                                 host_name = deprecated(),
                                 host_ip = deprecated(),
                                 port = deprecated()) {
-
```
hadley, line 109: Could you revert these whitespace changes please?

R/oauth-flow-password.R
```diff
@@ -69,9 +69,20 @@ oauth_flow_password <- function(client,
...
+    result <- rstudioapi::askForPassword(prompt)
+  } else {
```
hadley, line 82: Should this have an `!is_interactive()` branch with an informative error?

hadley:
Thanks for working on this!

R/oauth-flow-auth-code.R
```diff
@@ -403,23 +403,10 @@ oauth_flow_auth_code_pkce <- function() {
   )
 }
 
-# Try to determine whether we can redirect the user's browser to a server on
-# localhost, which isn't possible if we are running on a hosted platform.
-#
-# Currently this detects RStudio Server, Posit Workbench, and Google Colab. It
-# is based on the strategy pioneered by the {gargle} package.
-is_hosted_session <- function() {
```
fh-mthomson, line 411: moved to `utils.R` to:
1. Live with `is_rstudio_session()` function (new helper for `prompt_user()`)
2. Reflect usage beyond only `oauth-flow-auth-code` (#413)

R/oauth-flow-password.R
fh-mthomson: I don't _think_ it's needed?
* Within RStudio: `rstudioapi::askForPassword()` still works (surprising, but very welcome) when knitting
image
* If running outside of RStudio (e.g., Connect) then `is_rstudio_session()` is already false

Are there other cases I might be overlooking?

R/oauth-flow-auth-code.R
hadley: Perfect, thanks.

vignettes/articles/wrapping-apis.Rmd
hadley, line 510:
```suggestion
Using rstudioapi (or similar) here is good practice since you don't want to encourage the user to type their secret key into the console, as mentioned above.
```

R/oauth-flow-password.R
hadley: There are two scenarios running outside of RStudio that I'm worried about:

* You're running R interactively and call `knit()`: does `readline()` cause execution to pause? Do you see the correct prompt?
* You're running R in batch mode and call `knit()`: won't `readline()` will return `""` causing an authentication problem that's hard to diagnose?

hadley: Any more thoughts on this comment? I think it's the main blocker for merging this PR.

fh-mthomson: Sorry for the delay while juggling some other things. Of note, this PR is a specific workaround for Snowflake auth in Workbench and was before I chatted with @atheriel about his efforts to integrate OAuth more directly via Workbench (rather than our current wrappers around `httr2` which have been reasonably well).

Regardless, I had a chance to think a bit more about this, motivated by [Quarto's execution approach not supporting interactive user prompts](https://github.com/quarto-dev/quarto-cli/discussions/6076).

I've pushed a branch, per your suggestion and very helpful test cases! I _think_ it's closer to a reasonable approach.

Current behavior:

1. In RStudio: 
    a. .rmd -> knitting: pauses to receive prompt (new, yay!)
    b. .qmd -> render: fails with an expected, or at least, [well-documented](https://github.com/quarto-dev/quarto-cli/discussions/6076) error:
```
Error:
! RStudio not running
```
2. Outside of RStudio: both of your test cases:
```
## Error in `prompt_user()` at script.R:18:3:
## ! Unable to obtain user input in a non-interactive session.
```

Open to any thoughts!

fh-mthomson:
Going to close, for the [reasons above](https://github.com/r-lib/httr2/pull/410#discussion_r1527029756), namely that Workbench <> Snowflake authentication will be more robust for most use cases. 

Feel free to re-open if there's interest/benefits otherwise!

PR #417, opened 2023-12-15
Use list() and explicit as.character() for package versions

### LINKED ISSUE #416: invalid version specification ‘8.4.0-DEV’
This is coming from my curl:

```
curl::curl_version()$version
# [1] "8.4.0-DEV"
```

It seems to be coming from here:

https://github.com/r-lib/httr2/blob/ce918a6c83614cca402ebe2390c9bcc3bab7b0a7/R/req-options.R#L52C1-L56

Maybe `c()` is the wrong choice here? We shouldn't expect system libraries to fit R numeric versions, right?

MichaelChirico:
Closes #416


MichaelChirico:
I'd like to add a test with `local_mocked_bindings()`, but it's hampered by the current usage of namespaced calls (`utils::packageVersion()` and `curl::curl_version()`).

Which workaround would you prefer:

 1. `@importFrom utils packageVersion`, `@importFrom curl curl_version`
 2. helper e.g. `get_lib_versions()` to mock

I'll start with the latter for now.

MichaelChirico:
Hmm, `get_lib_versions()` would be uninteresting because it doesn't actually let us test "what happens when `curl_version()` gives something 'unexpected'". So I used `@importFrom` approach for now.

An intermediate alternative is a finer-grained wrapper, `curl_system_version = \() curl::curl_version()$version`. Happy to change to that instead.

R/req-options.R
```diff
@@ -49,12 +49,12 @@ req_user_agent <- function(req, string = NULL) {
...
-      httr2 = utils::packageVersion("httr2"),
-      `r-curl` = utils::packageVersion("curl"),
-      libcurl = curl::curl_version()$version
+    versions <- list(
+      httr2 = packageVersion("httr2"),
+      `r-curl` = packageVersion("curl"),
+      libcurl = curl_version()$version
     )
-    string <- paste0(names(versions), "/", versions, collapse = " ")
+    string <- paste0(names(versions), "/", vapply(versions, as.character, ""), collapse = " ")
```
hadley, line 57: I'd put the `as.character()` around the `packageVersion()` calls to make it more clear where the problem comes from.

hadley (commented):
I think it would be worth to to provide a `curl_system_version()` function because that would both avoid importing a function and make the mocking a bit simpler.

R/req-options.R
MichaelChirico: OK, switched

R/req-options.R
```diff
@@ -49,10 +49,10 @@ req_user_agent <- function(req, string = NULL) {
   check_request(req)
 
   if (is.null(string)) {
-    versions <- c(
-      httr2 = utils::packageVersion("httr2"),
-      `r-curl` = utils::packageVersion("curl"),
-      libcurl = curl::curl_version()$version
+    versions <- list(
```
MichaelChirico, line 52:
```suggestion
    versions <- c(
```

vapply() was also doing list->vector, if we have three character strings c() works again

hadley:
Thanks @MichaelChirico!

Issue #429, opened 2024-01-25
Minor performance enhancement to url_parse


DyfanJones:
Hi all,

I have been playing around with `httr2::url_parse` and I believe I have got some performance improvements. 

``` r
library(rlang)

# Original httr2 
# Note: used `httr2:::` to simplify reprex example
url_parse <- function(url) {
  httr2:::check_string(url)
  
  # https://datatracker.ietf.org/doc/html/rfc3986#appendix-B
  pieces <- parse_match(url, "^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\\?([^#]*))?(#(.*))?")
  
  scheme <- pieces[[2]]
  authority <- pieces[[4]]
  path <- pieces[[5]]
  query <- pieces[[7]]
  if (!is.null(query)) {
    query <- query_parse(query)
  }
  fragment <- pieces[[9]]
  
  # https://datatracker.ietf.org/doc/html/rfc3986#section-3.2
  pieces <- parse_match(authority %||% "", "^(([^@]+)@)?([^:]+)?(:([^#]+))?")
  
  userinfo <- pieces[[2]]
  if (!is.null(userinfo)) {
    if (grepl(":", userinfo)) {
      userinfo <- httr2:::parse_in_half(userinfo, ":")
    } else {
      userinfo <- list(userinfo, NULL)
    }
  }
  hostname <- pieces[[3]]
  port <- pieces[[5]]
  
  structure(
    list(
      scheme = scheme,
      hostname = hostname,
      username = userinfo[[1]],
      password = userinfo[[2]],
      port = port,
      path = path,
      query = query,
      fragment = fragment
    ),
    class = "httr2_url"
  )
}

# remove loop when applying NULL to empty elements
parse_match <- function(x, pattern) {
  m <- regexec(pattern, x, perl = TRUE)
  pieces <- regmatches(x, m)[[1]][-1]
  
  # replace empty element with null
  pieces[pieces == ""] <- list(NULL)
  return(pieces)
}

# same as original httr2
query_parse <- function(x) {
  x <- gsub("^\\?", "", x) # strip leading ?, if present
  params <- parse_name_equals_value(httr2:::parse_delim(x, "&"))
  
  if (length(params) == 0) {
    return(NULL)
  }
  
  out <- as.list(curl::curl_unescape(params))
  names(out) <- curl::curl_unescape(names(params))
  out
}

# use matrix operations instead
parse_name_equals_value <- function(x) {
[...]

PR #430, opened 2024-01-25
minor performance improvements to url_parse (#429)

DyfanJones:
This is a minor performance enhancement to function `url_parse`

Issue: https://github.com/r-lib/httr2/issues/429

``` r
url <- 'https://someurl.com/with/query_string?i=main&mode=front&sid=12ab&enc=+Hello'

(bm <- bench::mark(
  httr = httr::parse_url(url),
  httr2 = httr2::url_parse(url),
  new_mthd = url_parse(url), # new method in PR
  urltools = urltools::url_parse(url),
  adaR = adaR::ada_url_parse(url),
  check = F
))
#> # A tibble: 5 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 httr          239µs    256µs     3774.  387.71KB     35.5
#> 2 httr2         207µs    232µs     4100.  466.62KB     34.8
#> 3 new_mthd      129µs    144µs     6509.     4.5MB     39.2
#> 4 urltools      126µs    134µs     7024.    7.45MB     43.8
#> 5 adaR          189µs    202µs     4755.    1.06MB     47.1

bm |> ggplot2::autoplot()
#> Loading required namespace: tidyr
```

``` r

urls <- list(
  "/",
  "//google.com",
  "file:///",
  "http://google.com/",
  "http://google.com/path",
  "http://google.com/path?a=1&b=2",
  "http://google.com:80/path?a=1&b=2",
  "http://google.com:80/path?a=1&b=2#frag",
  "http://google.com:80/path?a=1&b=2&c=%7B1%7B2%7D3%7D#frag",
  "http://user@google.com:80/path?a=1&b=2",
  "http://user:pass@google.com:80/path?a=1&b=2",
  "svn+ssh://my.svn.server/repo/trunk",
  'https://someurl.com/with/query_string?i=main&mode=front&sid=12ab&enc=+Hello',
  "http://user:pass@example.com:80/path?a=1&b=2&c={1{2}3}#frag"
)

(bm <- bench::mark(
  httr = lapply(urls, httr::parse_url),
  httr2 = lapply(urls, httr2::url_parse),
  adaR = lapply(urls, adaR::ada_url_parse),
  urltools = lapply(urls, urltools::url_parse),
  new_mthd = lapply(urls, url_parse), # new method in PR
  check = F,
  iterations = 1000
))
#> # A tibble: 5 × 6
#>   expression      min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 httr          2.5ms   2.55ms      388.   161.9KB     18.3
#> 2 httr2        2.25ms   2.31ms      430.    70.9KB     20.7
#> 3 adaR         2.68ms   2.78ms      358.    34.9KB     24.5
#> 4 urltools     1.73ms    1.8ms      551.    34.9KB     23.5
#> 5 new_mthd     1.59ms   1.66ms      595.    62.9KB     22.2
bm |> ggplot2::autoplot()
```

<sup>Created on 2024-01-25 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/parse.R
+++ b/R/parse.R
@@ -1,7 +1,7 @@
 parse_media <- function(x) {
   # https://datatracker.ietf.org/doc/html/rfc2616#section-3.7
   pieces <- parse_delim(x, ";")
-  params <- parse_name_equals_value(pieces[-1])
+  params <- parse_name_equals_value(pieces[-1], ";")
 
   if (is_empty(pieces)) {
     list(type = NA_character_)
@@ -15,7 +15,7 @@ parse_www_authenticate <- function(x) {
   # https://stackoverflow.com/questions/10239970
 
   pieces <- parse_in_half(x, " ")
-  params <- parse_name_equals_value(parse_delim(pieces[[2]], ","))
+  params <- parse_name_equals_value(parse_delim(pieces[[2]], ","), ";")
 
   c(list(scheme = pieces[[1]]), params)
 }
@@ -27,7 +27,7 @@ parse_link <- function(x) {
     pieces <- parse_delim(x, ";")
 
     url <- gsub("^<|>$", "", pieces[[1]])
-    params <- parse_name_equals_value(pieces[-1])
+    params <- parse_name_equals_value(pieces[-1], ";")
     c(list(url = url), params)
   }
 
@@ -51,9 +51,14 @@ parse_delim <- function(x, delim, quote = "\"", ...) {
   )
 }
 
-parse_name_equals_value <- function(x) {
-  if (length(x) == 0) return(NULL)
-  pieces <- strsplit(x, "=")
+parse_name_equals_value <- function(x, safe = "&", quote = "\"") {
+  if (length(x) == 0) {
+    return(NULL)
+  }
+  # Note: quotes are removed in the parse_delim function
+  # re-introduce quotes to create a safe pattern to split on
+  safe_pattern <- paste0(quote, safe, quote)
+  pieces <- strsplit(sub("=", safe_pattern, x, fixed = T), safe_pattern, fixed = TRUE)
   pieces_matrix <- do.call(rbind, pieces)
 
   if (ncol(pieces_matrix) == 1) {
@@ -61,9 +66,9 @@ parse_name_equals_value <- function(x) {
   }
 
   # If only one piece, assume it's a field name with empty value
-  found <- pieces_matrix[,1] == pieces_matrix[,2]
+  found <- pieces_matrix[, 1] == pieces_matrix[, 2]
   pieces_matrix[found, 2] <- ""
-  set_names(as.list(pieces_matrix[,2]), pieces_matrix[,1])
+  set_names(as.list(pieces_matrix[, 2]), pieces_matrix[, 1])
 }
 
 parse_in_half <- function(x, pattern) {
@@ -79,6 +84,7 @@ parse_match <- function(x, pattern) {
   m <- regexec(pattern, x, perl = TRUE)
   pieces <- regmatches(x, m)[[1]][-1]
 
+  pieces <- as.list(pieces)
   # replace empty element with null
   pieces[pieces == ""] <- list(NULL)
   return(pieces)
@@ -87,4 +93,3 @@ parse_match <- function(x, pattern) {
 empty_to_null <- function(x) {
   if (x == "") NULL else x
 }
-
--- a/R/parse.R
+++ b/R/parse.R
@@ -52,17 +52,18 @@ parse_delim <- function(x, delim, quote = "\"", ...) {
 }
 
 parse_name_equals_value <- function(x) {
-  loc <- regexpr("=", x, fixed = TRUE)
-  pieces <- regmatches(x, loc, invert = TRUE)
+  if (length(x) == 0) return(NULL)
+  pieces <- strsplit(x, "=")
+  pieces_matrix <- do.call(rbind, pieces)
+
+  if (ncol(pieces_matrix) == 1) {
+    pieces_matrix <- cbind(pieces_matrix, rep("", nrow(pieces_matrix)))
+  }
 
   # If only one piece, assume it's a field name with empty value
-  expand <- function(x) if (length(x) == 1) c(x, "") else x
-  pieces <- map(pieces, expand)
-
-  val <- trimws(map_chr(pieces, "[[", 2))
-  name <- trimws(map_chr(pieces, "[[", 1))
-
-  set_names(as.list(val), name)
+  found <- pieces_matrix[,1] == pieces_matrix[,2]
+  pieces_matrix[found, 2] <- ""
+  set_names(as.list(pieces_matrix[,2]), pieces_matrix[,1])
 }
 
 parse_in_half <- function(x, pattern) {
@@ -77,7 +78,10 @@ parse_in_half <- function(x, pattern) {
 parse_match <- function(x, pattern) {
   m <- regexec(pattern, x, perl = TRUE)
   pieces <- regmatches(x, m)[[1]][-1]
-  map(pieces, empty_to_null)
+
+  # replace empty element with null
+  pieces[pieces == ""] <- list(NULL)
+  return(pieces)
 }
 
 empty_to_null <- function(x) {
--- a/R/parse.R
+++ b/R/parse.R
@@ -15,7 +15,7 @@ parse_www_authenticate <- function(x) {
   # https://stackoverflow.com/questions/10239970
 
   pieces <- parse_in_half(x, " ")
-  params <- parse_name_equals_value(parse_delim(pieces[[2]], ","), ";")
+  params <- parse_name_equals_value(parse_delim(pieces[[2]], ","), ",")
 
   c(list(scheme = pieces[[1]]), params)
 }
@@ -56,18 +56,29 @@ parse_name_equals_value <- function(x, safe = "&", quote = "\"") {
     return(NULL)
   }
   # Note: quotes are removed in the parse_delim function
-  # re-introduce quotes to create a safe pattern to split on
+  # - Re-introduce quotes to create a safe pattern to split on
+  # - Replaces the first instance of = with the safe pattern
+  # For example:
+  # c("var1=&", "var2==") -> c("var1\"&\"&", "var2\"&\"=")
   safe_pattern <- paste0(quote, safe, quote)
   pieces <- strsplit(sub("=", safe_pattern, x, fixed = T), safe_pattern, fixed = TRUE)
+
+  pieces_len <- lengths(pieces)
+  # If only one piece, assume it's a field name with empty value
+  found_single_field_piece <- pieces_len == 1
+
+  # replace empty pieces
+  # i.e. character(0) -> ""
+  pieces[pieces_len == 0] <- ""
+
+  # create matrix from nested list
   pieces_matrix <- do.call(rbind, pieces)
 
   if (ncol(pieces_matrix) == 1) {
     pieces_matrix <- cbind(pieces_matrix, rep("", nrow(pieces_matrix)))
   }
-
-  # If only one piece, assume it's a field name with empty value
-  found <- pieces_matrix[, 1] == pieces_matrix[, 2]
-  pieces_matrix[found, 2] <- ""
+  # replace single field piece with empty value
+  pieces_matrix[found_single_field_piece, 2] <- ""
   set_names(as.list(pieces_matrix[, 2]), pieces_matrix[, 1])
 }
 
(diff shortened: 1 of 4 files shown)


R/parse.R
```diff
@@ -77,7 +78,10 @@ parse_in_half <- function(x, pattern) {
 parse_match <- function(x, pattern) {
   m <- regexec(pattern, x, perl = TRUE)
   pieces <- regmatches(x, m)[[1]][-1]
-  map(pieces, empty_to_null)
+
+  # replace empty element with null
+  pieces[pieces == ""] <- list(NULL)
```
hadley, line 83: This is a bit confusing to me because it implicitly converts the character vector to a list. Maybe this would be clearer?

```R
empty <- pieces == ""
pieces <- as.list(pieces)
pieces[empty] <- list(NULL)
```

Or maybe use `ifelse()`?

```R
ifelse(pieces == "", list(NULL), as.list(pieces))
```

Would you mind seeing if those approaches work and what the impact on the benchmarks is?

R/parse.R
```diff
@@ -52,17 +52,18 @@ parse_delim <- function(x, delim, quote = "\"", ...) {
 }
 
 parse_name_equals_value <- function(x) {
-  loc <- regexpr("=", x, fixed = TRUE)
-  pieces <- regmatches(x, loc, invert = TRUE)
+  if (length(x) == 0) return(NULL)
+  pieces <- strsplit(x, "=")
```
hadley, line 56: What happens if there are multiple equals in the string? It doesn't look like that case is currently covered in the tests, so you might want to add a tests specifically for `parse_name_equals_value()` too.

hadley (commented):
Thanks for working on this!

Would you mind also including before and after benchmarks of the functions specifically affected here?

R/parse.R
DyfanJones: The only case it doesn't work is this type of scenario: `foo==&bar=test`. To overcome with we can do a few options:

Original Method:
```r
loc <- regexpr("=", x, fixed = TRUE)
regmatches(x, loc, invert = TRUE)
```

stringi method:
```r
stringi::stri_split_fixed(x, "=", n = 2)
```

substitution method:
```r
#i.e.
safe = "&"
strsplit(sub("=", safe, x, fixed = T), safe, fixed = TRUE)
```
**NOTE:** As `parse_name_equals_value` is called after `parse_delim` we can use the splitting delimiter in `parse_delim` as a "safe" character for the substitution method.


### Benchmark example:
``` r
safe <- "&"
query <- c(
  "foo==", "foo==&bar==", "foo==&bar==&cho==", "foo==&bar==&cho==&zoo=="
)
query_param <- lapply(query,
  httr2:::parse_delim, delim = safe
)

bms <- lapply(seq_along(query_param), \(i) {
  x <- query_param[[i]]
  bm <- bench::mark(
    original = {
      loc <- regexpr("=", x, fixed = TRUE)
      regmatches(x, loc, invert = TRUE)
    },
    stringi = stringi::stri_split_fixed(x, "=", n = 2),
    substitution = strsplit(sub("=", safe, x, fixed = T), safe, fixed = TRUE)
  )
  bm$param_no <- i
  bm
})

bms |> dplyr::bind_rows() |> ggplot2::autoplot()
#> Loading required namespace: tidyr
```



<sup>Created on 2024-01-26 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

The substitution method has some really good performance with it only being slightly slower than `stringi`.

DyfanJones:
Before and after benchmark

```r
urls <- list(
  "/",
  "//google.com",
  "file:///",
  "http://google.com/",
  "http://google.com/path",
  "http://google.com/path?a=1&b=2",
  "http://google.com:80/path?a=1&b=2",
  "http://google.com:80/path?a=1&b=2#frag",
  "http://google.com:80/path?a=1&b=2&c=%7B1%7B2%7D3%7D#frag",
  "http://user@google.com:80/path?a=1&b=2",
  "http://user:pass@google.com:80/path?a=1&b=2",
  "svn+ssh://my.svn.server/repo/trunk",
  'https://someurl.com/with/query_string?i=main&mode=front&sid=12ab&enc=+Hello',
  "http://user:pass@example.com:80/path?a=1&b=2&c={1{2}3}#frag"
)

bm <- bench::mark(
  lapply(urls, httr2::url_parse),
  iterations = 1000
)


#> # A tibble: 2 × 13
#>   expression      min   median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result      memory              time               gc                  
#>   <bch:expr> <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list>      <list>              <list>             <list>              
#> 1 before       2.24ms   2.32ms      421.    62.9KB     46.8   900   100      2.14s <list [14]> <Rprofmem [27 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
#> 2 after        1.65ms   1.72ms      573.    62.9KB     20.8   965    35      1.68s <list [14]> <Rprofmem [28 × 3]> <bench_tm [1,000]> <tibble [1,000 × 3]>
```

image

I am guessing there will be micro performance enchancements across httr2 as `parse_name_equals_value` is the main function that has been changed.

R/parse.R
DyfanJones:
```r
empty <- pieces == ""
pieces <- as.list(pieces)
pieces[empty] <- list(NULL)
```
Had a small (hardly noticeable) impact on performance to the implicit conversion of character vector to a list.


```r
ifelse(pieces == "", list(NULL), as.list(pieces))
```
Sadly had a larger impact and gave performance similar to the original implementation.

hadley: Yeah, lets go with the explicit coercion to a list then; I think that's easier to reason about.

R/parse.R
```diff
@@ -77,10 +83,14 @@ parse_in_half <- function(x, pattern) {
 parse_match <- function(x, pattern) {
   m <- regexec(pattern, x, perl = TRUE)
   pieces <- regmatches(x, m)[[1]][-1]
-  map(pieces, empty_to_null)
+
+  empty <- pieces == ""
+  pieces <- as.list(pieces)
+  # replace empty element with null
```
hadley, line 89: I think you can drop the comment as that's what the code expresses.

R/parse.R
```diff
@@ -51,18 +51,24 @@ parse_delim <- function(x, delim, quote = "\"", ...) {
...
 
-parse_name_equals_value <- function(x) {
-  loc <- regexpr("=", x, fixed = TRUE)
-  pieces <- regmatches(x, loc, invert = TRUE)
+parse_name_equals_value <- function(x, safe = "&", quote = "\"") {
+  if (length(x) == 0) {
+    return(NULL)
+  }
+  # Note: quotes are removed in the parse_delim function
+  # re-introduce quotes to create a safe pattern to split on
```
hadley, line 59: Why do we need quotes and `&` to create the safe pattern?

I think it would also be useful to add a comment pointing out that we're replacing the first instance of `=` with a sequence that we know doesn't exist in the string to ensure that we always produce at most two pieces.

R/parse.R
```diff
@@ -51,18 +51,24 @@ parse_delim <- function(x, delim, quote = "\"", ...) {
...
 
   # If only one piece, assume it's a field name with empty value
-  expand <- function(x) if (length(x) == 1) c(x, "") else x
-  pieces <- map(pieces, expand)
-
-  val <- trimws(map_chr(pieces, "[[", 2))
-  name <- trimws(map_chr(pieces, "[[", 1))
-
-  set_names(as.list(val), name)
+  found <- pieces_matrix[, 1] == pieces_matrix[, 2]
```
hadley, line 69: This happens because of recycling in `rbind()`?

hadley (commented):
Thanks for continuing to work on this 😄 

For the benchmarks, I meant can you please benchmark `parse_name_equals_value()` and `parse_match()` by themselves? Just so I get a better understanding of what was causing the slowness before.

DyfanJones:
Ah sorry about the benchmarks, my mistake:


``` r
parse_match <- function(x, pattern) {
  m <- regexec(pattern, x, perl = TRUE)
  pieces <- regmatches(x, m)[[1]][-1]
  empty <- pieces == ""
  pieces <- as.list(pieces)
  pieces[empty] <- list(NULL)
  return(pieces)
}

urls <- list(
  "/",
  "//google.com",
  "file:///",
  "http://google.com/",
  "http://google.com/path",
  "http://google.com/path?a=1&b=2",
  "http://google.com:80/path?a=1&b=2",
  "http://google.com:80/path?a=1&b=2#frag",
  "http://google.com:80/path?a=1&b=2&c=%7B1%7B2%7D3%7D#frag",
  "http://user@google.com:80/path?a=1&b=2",
  "http://user:pass@google.com:80/path?a=1&b=2",
  "svn+ssh://my.svn.server/repo/trunk",
  'https://someurl.com/with/query_string?i=main&mode=front&sid=12ab&enc=+Hello',
  "http://user:pass@example.com:80/path?a=1&b=2&c={1{2}3}#frag"
)
pattern <- "^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\\?([^#]*))?(#(.*))?"
bench::mark(
  before = lapply(urls, httr2:::parse_match, pattern = pattern),
  after = lapply(urls, parse_match, pattern = pattern)
) |> ggplot2::autoplot() + ggplot2::labs(title = "parse_match benchmark")
#> Loading required namespace: tidyr
```
image

<sup>Created on 2024-01-29 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

`parse_match` has a micro performance improvement.

``` r
library(rlang)
parse_name_equals_value <- function(x, safe = "&", quote = "\"") {
  if (length(x) == 0) {
    return(NULL)
  }
  # Note: quotes are removed in the parse_delim function
  # re-introduce quotes to create a safe pattern to split on
  safe_pattern <- paste0(quote, safe, quote)
  pieces <- strsplit(sub("=", safe_pattern, x, fixed = T), safe_pattern, fixed = TRUE)
  pieces_matrix <- do.call(rbind, pieces)
  
  if (ncol(pieces_matrix) == 1) {
    pieces_matrix <- cbind(pieces_matrix, rep("", nrow(pieces_matrix)))
  }
  
  # If only one piece, assume it's a field name with empty value
  found <- pieces_matrix[, 1] == pieces_matrix[, 2]
  pieces_matrix[found, 2] <- ""
  set_names(as.list(pieces_matrix[, 2]), pieces_matrix[, 1])
}

params <- "foo==&var1=12ab,var2=;var3=,&var4=\"&\"var5=;out"
bench::mark(
  before = httr2:::parse_name_equals_value(httr2:::parse_delim(params, "&")),
  after = parse_name_equals_value(httr2:::parse_delim(params, "&"), "&")
) |> ggplot2::autoplot() + ggplot2::labs(title = "parse_name_equals_value benchmark")
#> Loading required namespace: tidyr
```

image

<sup>Created on 2024-01-29 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

R/parse.R
DyfanJones: Yes for example:

``` r
ll <- list(
  c("var1", "foo"),
  c("var2"),
  c("var3", "bar")
)

do.call(rbind, ll)
#>      [,1]   [,2]  
#> [1,] "var1" "foo" 
#> [2,] "var2" "var2"
#> [3,] "var3" "bar"
```

<sup>Created on 2024-01-29 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

DyfanJones:
@hadley I have added in some extra comments and I have restructured `parse_name_equals_value` function. I am hoping it makes more sense in the logic it uses (note these changes have no impact on the current performance).

R/parse.R
```diff
@@ -51,18 +51,35 @@ parse_delim <- function(x, delim, quote = "\"", ...) {
   )
 }
 
-parse_name_equals_value <- function(x) {
-  loc <- regexpr("=", x, fixed = TRUE)
-  pieces <- regmatches(x, loc, invert = TRUE)
-
+parse_name_equals_value <- function(x, safe = "&", quote = "\"") {
```
hadley, line 54: I think I've figured out a faster way to do this which is also a little clearer (if more verbose): https://gist.github.com/hadley/2fd32cdc053099d62d56f5f28898ec95

The main insight I had is that main culprit for the slow speed in my approach is `regmatches()`, and for this simple case, it's easy enough to do by hand. I also named the function `parse_in_half()` as there's an existing function with that name that only works for scalars.

DyfanJones: Plus it looks like you get around the R magic from the `do.call(rbind, nested_list)`.

PR #444, opened 2024-02-16
Docs fix

jameslairdsmith:
I think this is correct. It's the response body that is being saved, not the request body.


hadley:
Oops yes, thanks for catching!

PR #473, opened 2024-06-01
Fix `@return` value doc

salim-b:


hadley:
I think you have to be an admin for those helpers to work

salim-b:
> I think you have to be an admin for those helpers to work

I hoped it would also work the PR author, but no 😐

PR #492, opened 2024-07-11
Handle streaming errors like regular errors

### LINKED ISSUE #479: Cannot capture 4xx response bodies with req_perform_stream
I'm finding a difference in behavior between `req_perform` and `req_perform_stream` when it comes to capturing HTTP errors and translating them to R errors, and moreover in capturing 4xx response bodies. Both occur straightforwardly with `req_perform`, following the documented instructions:

```r
req <- ...
req <- req_error(req, body = \(resp) resp_body_string(resp)
resp <- req_perform(req)
```
This will throw an R error that includes the text body from the error message.

However, error responses are not handled and converted to R errors in the same way using `req_perform_stream`:
```r
> req <- ...
> req <- req_error(req, body = \(resp) resp_body_string(resp)
> req_perform(req)
<httr2_response>
POST http://obfuscated
Status: 400 Bad Request
Content-Type: text/plain
Body: None
```
This returns a response object with an error status code is returned, but does not throw an R error. It also does not include any body. This is true even when setting `is_error` to a lambda returning FALSE. In the `req_perform` case, that successfully turns off the http-to-R error handling and yields a 4xx response with a body, but for `req_perform_stream` the response body is still empty. The following obfuscated example uses a private API and so is not a reprex:

```r
> request <- ...
> request <- httr2::req_error(request, is_error = \(z) FALSE)
> httr2::req_perform(request)
<httr2_response>
POST http://obfuscated
Status: 400 Bad Request
Content-Type: text/plain
Body: In memory (147 bytes)

> httr2::req_perform_stream(request, cb, buffer_kb = buffer_kb)
<httr2_response>
POST http://obfuscated
Status: 400 Bad Request
Content-Type: text/plain
Body: None
```

My current workaround is to repeat the request using `req_perform` when `req_perform_stream` returns an error status, but this can mean repeating large POST requests. Is there a way to get the 400 response's along with the single `req_perform_stream` request?

hadley:
Fixes #479

@jwimberl how does this look to you?

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httr2 (development version)
 
+* `req_perform_stream()` no longer applies the `callback` to unsuccessful
+  responses, instead creating a regular response. It also now sets `last_request()` and `last_response()` (#479).
 * `req_body_file()` now only opens a connection when the request actually needs data. In particular, this makes it work better with `req_perform_parallel()` (#487).
 * `req_perform_parallel()` now works when downloading 0 byte files (#478)
 * Corrupt `rds` files no longer cause the request to fail.
--- a/R/req-perform-stream.R
+++ b/R/req-perform-stream.R
@@ -1,9 +1,14 @@
 
 #' Perform a request and handle data as it streams back
-#'
+#' 
+#' @description
 #' After preparing a request, call `req_perform_stream()` to perform the request
 #' and handle the result with a streaming callback. This is useful for
 #' streaming HTTP APIs where potentially the stream never ends.
+#' 
+#' The `callback` will only be called if the result is successful. If you need 
+#' to stream an error response, you can use [req_error()] to suppress error
+#' handling so that the body is streamed to you.
 #'
 #' @inheritParams req_perform
 #' @param callback A single argument callback function. It will be called
@@ -15,7 +20,9 @@
 #'   Choose `"byte"`, `"line"`, or supply your own function that takes a
 #'   raw vector of `bytes` and returns the locations of possible cut points
 #'   (or `integer()` if there are none).
-#' @returns An HTTP [response].
+#' @returns An HTTP [response]. The body will be empty if the request was
+#'   successful (since the `callback` function will have handled it). The body
+#'   will contain the HTTP response body if the request was unsuccessful.
 #' @export
 #' @examples
 #' show_bytes <- function(x) {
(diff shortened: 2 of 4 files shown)


jwimberl:
Testing this branch, I see uniform behavior btw req_perform and req_perform_stream in one of the reprex cases from #479:

``` r
packageVersion("httr2")
#> [1] '1.0.1.9000'
library(httr2)
req <- request("https://api.github.com/asdfsdadf")
a <- req |> req_perform()
#> Error in `req_perform()`:
#> ! HTTP 404 Not Found.
last_response() |> resp_body_string()
#> [1] "{\n  \"message\": \"Not Found\",\n  \"documentation_url\": \"https://docs.github.com/rest\",\n  \"status\": \"404\"\n}\n"
b <- req |> req_perform_stream(callback = \(z) TRUE)
#> Error in `req_perform_stream()`:
#> ! HTTP 404 Not Found.
last_response() |> resp_body_string()
#> [1] "{\n  \"message\": \"Not Found\",\n  \"documentation_url\": \"https://docs.github.com/rest\",\n  \"status\": \"404\"\n}\n"
```

<sup>Created on 2024-07-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

So, both lead to an R error, and `last_response()` contains the error body in both cases, which looks good to me, as this most likely how I would want to call endpoints like this.

I do see a slight difference behavior in the other reprex case:

``` r
packageVersion("httr2")
#> [1] '1.0.1.9000'
library(httr2)
req <- request("https://api.github.com/asdfsdadf")
a <- req |> req_error(is_error = \(z) FALSE) |> req_perform()
last_response() |> resp_body_string()
#> [1] "{\n  \"message\": \"Not Found\",\n  \"documentation_url\": \"https://docs.github.com/rest\",\n  \"status\": \"404\"\n}\n"
b <- req |> req_error(is_error = \(z) FALSE) |> req_perform_stream(callback = \(z) TRUE)
last_response() |> resp_body_string()
#> Error in `resp_body_raw()`:
#> ! Can't retrieve empty body.
```

<sup>Created on 2024-07-12 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>


but this makes sense too -- if I'm overriding things to tell it its not an error, of course it would leave the body to be handled by the callback. So, this seems consistent to me as well.

hadley:
@jwimberl thanks for the thoughtful review. I really appreciate it 😄

PR #499, opened 2024-07-11
Check password more lazily

### LINKED ISSUE #498: `req_oauth_password` prompts password-input continuously, but existing valid token are used
I was facing some issues in #491, but now the API part are working as expected, so I'm able to use `req_oauth_password`. However, I noticed that #119 points out that `req_oauth_password` triggers the "Password" input prompt on **all** request. Even though, it's actually saving the token in the first execution, and **does** re-use the token in the next execution.

Have I implemented something in a wrong way for my first "initial httr2-based test project"? There's no solution in #119, but [hadley](https://github.com/hadley) does request an example.

I use the following simple test-code to try and fetch "me" from "users":

```r
get_users_me <- function() {
  response <- NULL

  if (interactive()) {
    tryCatch({
      response <- request("https://domain.com/users/me") |>
       req_auth() |>
       req_perform()
     
     # Handle the response inside the conditional
     content <- response |> resp_body_json(auto_unbox = TRUE)
    }, error = function(e) {
      stop("An error occurred: ", conditionMessage(e))
    })
  } else {
    # Handle non-interactive case, e.g., tests or scripts
    # Make sure to define how response should be handled or mocked here
  }

  # return content
  content
}
```

I use the following for my (currently) test client, and the wrapper for my req_auth:

```r
test_con_client <- function() {
  oauth_client(
    id = "28acfec0674bb3da9f39",
    token_url = "https://domain.com/oauth/token",
    name = "client-r-package"
  )
}

req_auth <- function(req) {
  req_oauth_password(req,
    client = test_con_client(),
    username = "user@domain.com"
  )
}
```

Using `cache_disk` does not change a thing. But it do look like it's using the existing token, even through it re-prompts me, asking me to enter the password again.

I noticed this by running `with_verbosity(get_users_me())`:

1) `with_verbosity(get_users_me())`
 a) does a `POST /oauth/token HTTP/2`
 b) get's a `HTTP/2 200`
 c) requests the resources at `GET /users/me HTTP/2` (The Bearer are set in the headers, as expected)
 d) gets a `HTTP/2 200`

1) _same R session, `with_verbosity(get_users_me())` executed again._
  a) does a `GET /users/me HTTP/2`  (The Bearer are set in the headers, as expected)
  b) gets a `HTTP/2 200`

However, the password input prompt shows up on both executions:

billede

_Ignore the code in the background._

In other words, it's clear to me, that the password are not used in run no. 2, but are still being prompted to the user, for some reason.

Is this a bug, or do **I** need to check something, and only wrap `req_oauth_password`, if it needs to be re-authenticated? (initial, once expired, and so on)?

If I run the official [example with github_client](https://httr2.r-lib.org/articles/oauth.html#authenticating-a-request) with [`req_oauth_auth_code`](https://httr2.r-lib.org/reference/req_oauth_auth_code.html), things works as I expect it to do: 1. run prompts me to grant access, and fetch the information. 2. run just fetches the information. I'm not "reprompted" to grant access.

hadley:
Maybe fixes #498

@exetico I don't have any way to debug this because I don't have access to an API that uses this style of auth, but I changed two lines that looked suspicious when I took a second look.

### DIFF
--- a/R/oauth-flow-password.R
+++ b/R/oauth-flow-password.R
@@ -33,7 +33,6 @@ req_oauth_password <- function(req,
                                cache_disk = FALSE,
                                cache_key = username) {
 
-  password <- check_password(password)
   params <- list(
     client = client,
     username = username,
@@ -56,12 +55,11 @@ oauth_flow_password <- function(client,
     interactive = is.null(password)
   )
   check_string(username)
-  password <- check_password(password)
 
   oauth_client_get_token(client,
     grant_type = "password",
     username = username,
-    password = password,
+    password = check_password(password),
     scope = scope,
     !!!token_params
   )


exetico:
I've executed `install_github('https://github.com/r-lib/httr2/tree/password-prompts')`, replaced all instances of the package and restarted my R session.

Now I'm not getting the 2nd. popup. That was a easy fix, huh 😄.

To show my gratitude, I've also spinned up my local API instance, and bumped down the expire_in to 20 sec, and after that, it's reprompting me, as expected.

Thank you!

I've read the policy of "how often to release updates to R-packages", and I can see that it's a good amount of time ago, since the last release.

Do you have any idea on when this can be released (and available on CRAN)?

hadley:
I'm planning on a release next week

hadley:
Thanks so much for double checking!

PR #505, opened 2024-07-15
Add function req_perform_promise (fixes #501)

### LINKED ISSUE #501: feature request: `req_perform_promise()`
It's nice that parallelization is possible through `req_perform_multi()` but I wish there was a way that httr2 wouldn't have to hold onto control while the requests are being made. (For example, a shiny dashboard where each tile could load independently even if each depended on a httr2 request).

Playing around a bit, I was pleasantly surprised how easy it was to extend the internal `Performance` class to get a promise, based on the way the `promises` package wraps `future` objects (with a `later` loop), but that class isn't exported. Would you consider a PR that adds this? Or if not, exporting the `Performance` object?

gergness:
An implementation of `req_perform_promise()` - #501 

It has worked pretty well in my initial testing, though the way I'm tracking which pools already have a later task polling them does feel a little fragile.


R/req-promise.R
```diff
@@ -0,0 +1,156 @@
...
+#'
+#' @inheritParams req_perform
+#' @inheritParams req_perform_parallel
+#'
+#' @return a [promises::promise()] object which resolves to a [response] if
+#' successful or rejects on the same errors thrown by [req_perform()].
+#' @export
+#'
+#' @examples
+#' if (FALSE) { # Don't run
```
hadley, line 32:
```suggestion
#' @examplesIf requireNamespace("promises", quietly = TRUE)
```

?

R/req-promise.R
```diff
@@ -0,0 +1,156 @@
...
+        # cached
+        self$resolve(self$resp)
+        return()
+      }
+      super$submit()
+      ensure_pool_poller(pool, self$reject)
+    },
+
+    succeed = function(res) {
+      try_fetch({
```
hadley, line 109: Can you use `tryCatch()` or `withCallingHandlers()` here?

hadley (commented):
I added a couple of small comments but I'll try and find someone who knows more about promises than me to take a deeper look.

R/req-promise.R
```diff
@@ -0,0 +1,154 @@
...
+    },
+
+    succeed = function(res) {
+      tryCatch({
+        super$succeed(res)
+        self$resolve(self$resp)
+      },
+        httr2_fail = function(cnd) self$reject(cnd$error),
+        error = function(cnd) self$reject(cnd)
+      )
```
hadley, line 113: Would you mind using this indenting style? I find it a bit easier to read.

```suggestion
      tryCatch(
        {
          super$succeed(res)
          self$resolve(self$resp)
        },
        httr2_fail = function(cnd) self$reject(cnd$error),
        error = function(cnd) self$reject(cnd)
      )
```

jcheng5 (approved):
Impressive work, thank you!

PR #506, opened 2024-07-15
req_perform_parallel respects req_error settings

gergness:
Discovered while working on #505 - It seems like an oversight that `req_perform_multi()` doesn't respect `req_error()` error handling.


hadley (approved):
Looks great, thanks! I'm in the middle of a release right now so I'll merge when its done.

PR #514, opened 2024-08-11
Improving caching logic

hadley:
I can't reproduce the failures locally, but hopefully this fixes the problems found on https://cran.r-project.org/web/checks/check_results_pkgdown.html.


gaborcsardi (approved):
Yes, I agree that this is the right approach, you have to make cache reads "atomic".

PR #520, opened 2024-08-27
Allow `req_perform_stream()` to perform non-blocking reads

### LINKED ISSUE #519: Allow `req_perform_stream()` to specify max wait time
So it calls the callback either when the buffer is full or when the time has elapsed.

hadley:
The default behaviour stays the same, but you can now specify `wait_for` in order to switch to a non-blocking request and call the callback as the data arrives.

Fixes #519

@jcheng5 Does that feel like the right argument name? I have kept the behaviour the same as previously, but maybe the default should be `wait_for = 0`? Does the `wait_for` implementation loop look ok to you?

### DIFF
--- a/R/req-perform-stream.R
+++ b/R/req-perform-stream.R
@@ -15,6 +15,11 @@
 #'   worth of data to process. It must return `TRUE` to continue streaming.
 #' @param timeout_sec Number of seconds to process stream for.
 #' @param buffer_kb Buffer size, in kilobytes.
+#' @param wait_for Number of seconds to wait for more data before calling
+#'   `callback`. If this is `Inf` (the default), `callback` will be called
+#'   whenever the buffer is full. If this is `0`, `callback` will be
+#'   called whenever there is any data. For other values, `callback` will be
+#'   called whenever the buffer is full or `wait_for` seconds have passed.
 #' @param round How should the raw vector sent to `callback` be rounded?
 #'   Choose `"byte"`, `"line"`, or supply your own function that takes a
 #'   raw vector of `bytes` and returns the locations of possible cut points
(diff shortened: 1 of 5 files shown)


R/req-perform-stream.R
```diff
@@ -105,7 +126,11 @@ as_round_function <- function(round = c("byte", "line"),
     round <- arg_match(round, error_call = error_call)
     switch(round,
       byte = function(bytes) length(bytes),
-      line = function(bytes) which(bytes == charToRaw("\n"))
+      line = function(bytes) which(bytes == charToRaw("\n")),
+      sse  = function(bytes) {
+        is_newline <- which(bytes == charToRaw("\n"))
+        intersect(is_newline, is_newline + 1)
```
jcheng5, line 132: That's clever

jcheng5:
Have you tried this with a real SSE server and do you end up with an extra callback with "\n" at the end? If "yes" and "no", LGTM!

jcheng5 (approved):

tests/testthat/test-req-perform-stream.R
```diff
@@ -59,6 +59,37 @@ test_that("can buffer to lines", {
...
+  accumulate_bytes <- function(x) {
+    bytes <<- c(bytes, length(x))
+    TRUE
+  }
+
+  resp <- request_test("/stream-bytes/102400") %>%
+    req_url_query(chunk_size = 1024) %>%
+    req_perform_stream(accumulate_bytes, wait_for = 0)
+  expect_equal(sum(bytes), 102400)
+  # I'm not sure why this is so much smaller than 100, but I suspect it
```
hadley, line 73: @jcheng5 does this explanation make sense to you

PR #523, opened 2024-08-29
Clarify `max_tries` argument

hadley:


jonthegeek (approved):
I kept thinking you had it right before because of the 0 index, but nope, this fix is correct! I had to read the code a few times to get my brain to agree with that.

hadley:
@jonthegeek I convinced myself that the existing behaviour is correct, it's just that you never want `max_tries = 1` because 1 try = 0 retries. In hindsight, it would have been better to call this argument `max_retries`.

PR #533, opened 2024-09-03
Handle errors properly when `req_perform_connection(mode="text")`

jcheng5:
Fixes #532

PR #535, opened 2024-09-04
Manage our own pushback state

### LINKED ISSUE #532: `req_perform_stream(mode="text")` error during error handling
```r
library(httr2)

req <- request("https://api.openai.com/v1") |>
  req_auth_bearer_token("bad-openai-key") |>
  req_retry(max_tries = 2) |>
  req_error(body = function(resp) {
    httr2::resp_body_json(resp)$error$message
  }) |>
  req_url_path_append("/chat/completions") |>
  req_body_json(list(list(role = "user", content = "Why is the sky blue?"))) |>
  req_perform_connection(mode = "text")
```

Results in:

```
Error in `readBin()`:
! can only read from a binary connection
```

rather than the expected:

```
Error in `req_perform_connection()`:
! HTTP 401 Unauthorized.
• Incorrect API key provided: bad-open**-key. You can find your API key at [https://platform.openai.com/account/api-keys.](vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html#)
```

This is due to the `read_con` call in `req_perform_connection`, it's expecting a binary connection instead of a text connection.

hadley:
Fixes #532. Closes #533.

### DIFF
--- a/R/req-perform-stream.R
+++ b/R/req-perform-stream.R
@@ -98,7 +98,6 @@ req_perform_stream <- function(req,
 #' want to do other work in between handling inputs from the stream.
 #'
 #' @inheritParams req_perform_stream
-#' @param mode The mode that should be used for opening the connection.
 #' @param blocking When retrieving data, should the connection block and wait
 #'   for the desired information or immediately return what it has (possibly
 #'   nothing)?
@@ -115,13 +114,9 @@ req_perform_stream <- function(req,
 #'
 #' # Always close the response when you're done
 #' close(resp)
-req_perform_connection <- function(req,
-                                   mode = c("binary", "text"),
-                                   blocking = TRUE) {
+req_perform_connection <- function(req, blocking = TRUE) {
   check_request(req)
   check_bool(blocking)
-  mode <- arg_match(mode)
-  con_mode <- if (mode == "text") "rf" else "rbf"
 
   handle <- req_handle(req)
   the$last_request <- req
@@ -137,7 +132,7 @@ req_perform_connection <- function(req,
     if (!is.null(resp)) {
       close(resp$body)
     }
-    resp <- req_perform_connection1(req, handle, con_mode, blocking = blocking)
+    resp <- req_perform_connection1(req, handle, blocking = blocking)
 
     if (retry_is_transient(req, resp)) {
       tries <- tries + 1
@@ -159,11 +154,11 @@ req_perform_connection <- function(req,
   resp
 }
 
-req_perform_connection1 <- function(req, handle, con_mode = "rbf", blocking = TRUE) {
+req_perform_connection1 <- function(req, handle, blocking = TRUE) {
   stream <- curl::curl(req$url, handle = handle)
 
   # Must open the stream in order to initiate the connection
-  open(stream, con_mode, blocking = blocking)
+  open(stream,  "rbf", blocking = blocking)
   curl_data <- curl::handle_data(handle)
 
   new_response(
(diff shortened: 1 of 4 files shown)


jcheng5:
I think there's a bug in the way I used `pushBack` to begin with, and persists in this PR. When `blocking=FALSE`, `readLines` will return partial lines, and AFAICT we can't distinguish between partial and complete lines. I just pushed a unit test for this.

jcheng5:
Alright, I think this is ready

R/req-perform-stream.R
```diff
@@ -214,39 +209,116 @@ resp_stream_lines <- function(resp, lines = 1) {
...
+#     remaining = charToRaw("id: 12345")
+#   )
+find_event_boundary <- function(buffer) {
+  if (length(buffer) < 2) {
+    return(NULL)
+  }
+
+  # leftX means look behind by X bytes. For example, left1[2] equals buffer[1].
+  # Any attempt to read past the beginning of the buffer results in 0x00.
+  left1 <- c(0x00, head(buffer, -1))
```
hadley, line 252: Oh this is nice.

hadley:
Should we apply the same logic to `resp_stream_lines()` so that you always get a complete line? Or maybe move the logic that there so that `resp_stream_sse()` can use `resp_stream_lines()`?

R/req-perform-stream.R
```diff
@@ -214,39 +209,116 @@ resp_stream_lines <- function(resp, lines = 1) {
...
+  # Return a list with the event data and the remaining buffer
+  list(
+    matched = slice(buffer, end = split_at),
+    remaining = slice(buffer, start = split_at)
+  )
+}
+
 #' @export
 #' @rdname resp_stream_raw
 # TODO: max_size
```
hadley, line 278: Maybe we can implement `max_size()` now that we have `buffer`. But what should happen if `max_size` is reached? Error? Return `NULL`? Some other sentinel?

R/req-perform-stream.R
```diff
@@ -214,39 +209,116 @@ resp_stream_lines <- function(resp, lines = 1) {
...
-      return(parse_event(lines))
+    
+    # We didn't have enough data. Attempt to read more
+    chunk_size <- 1024
+    chunk <- readBin(resp$body, raw(), n = chunk_size)
+
+    print_buffer(chunk, "Received chunk")
+
+    # If we've reached the end of input, store the buffer and return NULL
+    if (length(chunk) == 0) {
```
hadley, line 311: Should this be `length(chunk) < chunk_size`?

R/req-perform-stream.R
```diff
@@ -214,39 +209,116 @@ resp_stream_lines <- function(resp, lines = 1) {
...
+  buffer <- resp$cache$push_back %||% raw()
+  resp$cache$push_back <- raw()
+
+  print_buffer <- function(buf, label) {
+    # cat(label, ":", paste(sprintf("%02X", as.integer(buf)), collapse = " "), "\n", file = stderr())
   }
 
-  lines <- character(0)
+  # Read chunks until we find an event or reach the end of input
   while (TRUE) {
```
hadley, line 291:
```suggestion
repeat {
```

R/req-perform-stream.R
jcheng5: Error for sure, IMHO. I'll add that

jcheng5:
> Should we apply the same logic to `resp_stream_lines()` so that you always get a complete line? Or maybe move the logic that there so that `resp_stream_sse()` can use `resp_stream_lines()`?

Uh... it looks like `readLines` already has the correct behavior--it only returns complete lines, even in nonblocking mode. Was all that logic in `resp_stream_sse` for nothing!? 😞

> And does this mean we're drifting away from the idea of blocking vs non-blocking? Or maybe that distinction is no longer useful?

No, I don't think so? I think we still need blocking for convenience, and non-blocking for async to be possible.

R/req-perform-stream.R
jcheng5: Now implemented

jcheng5:
Sorry, I was wrong about `readLines` not returning partial lines. So pushback would be good here.

jcheng5:
Well, that turned out to be quite an adventure. It's good to go now, AFAICT.

jcheng5 (approved):

jcheng5:
Sorry that my tests added seconds of latency to the test suite. Is there a way to nudge webfaker from the main testthat process, so we could get rid of those sleep statements?

PR #550, opened 2024-09-16
Validation for `url_build()` 

### LINKED ISSUE #482: url_build() doesn't validate argument
It's quite clear I didn't perfectly read the docs! But it would be quite nice if invalid names are found that an error (and better and informative one!) are provided. 

```r
httr2::url_build(list(host = host, port = port))
#> "//:3000"
```

Whereas I expected: 

```
#> "http://127.0.0.1:3000"
```

JosiahParry:
Closes #482 

This PR introduces validation logic to `url_build()`. 

In the event that an incorrectly named element is provided, an informative error message is provided to the user.


hadley:
Hmmm, it feels like the right validation here is that you're providing a httr2 url object. I'll implement that instead.

PR #580, opened 2024-11-11
Implements event-driven async for req_perform_promise()

### LINKED ISSUE #579: Use event-driven async for `req_perform_promise()`
[`later::later_fd()`](https://r-lib.github.io/later/reference/later_fd.html) provides the ability to wait efficiently on curl socket activity.

We update `req_perform_promise()` to use the new capability rather than poll for updates at 0.1 sec intervals.

A PR will be provided.

There will be separate PRs here and/or at `elmer` for streaming async.

shikokuchuo:
Instead of polling at 0.1 second intervals, we use `later::later_fd()` to wait upon the curl sockets efficiently. This triggers only when one of the sockets has activity that requires processing.

`later::loop_empty()` takes into account these file descriptor waits, hence there are no changes to the existing tests.

Additional tests have been added by @jcheng5 for the use of private event loops, which require a new pool to be created inside the loop.

This PR has already been reviewed and signed off by @jcheng5 over at my own fork: https://github.com/shikokuchuo/httr2/pull/1 There is a bit more discussion there, but mostly on the private loops, which is a fairly niche topic.

Requires dev `later` until the new version reaches CRAN.

Closes #579.

### DIFF
--- a/R/req-promise.R
+++ b/R/req-promise.R
@@ -142,8 +142,7 @@ ensure_pool_poller <- function(pool, reject) {
             readfds = fds$reads,
             writefds = fds$writes,
             exceptfds = fds$exceptions,
-            timeout = fds$timeout,
-            loop = later::global_loop()
+            timeout = fds$timeout
           )
         } else {
           monitor$ending()
--- a/R/req-promise.R
+++ b/R/req-promise.R
@@ -9,6 +9,10 @@
 #' [promises package documentation](https://rstudio.github.io/promises/articles/promises_01_motivation.html)
 #' for more details on how to work with the resulting promise object.
 #'
+#' If using together with [later::with_temp_loop()] or other private event loops,
+#' a new curl pool made by [curl::new_pool()] should be created for requests made
+#' within the loop to ensure that only these requests are being polled by the loop.
+#'
 #' Like with [req_perform_parallel()], exercise caution when using this function;
 #' it's easy to pummel a server with many simultaneous requests. Also, not all servers
 #' can handle more than 1 request at a time, so the responses may still return
--- a/R/req-promise.R
+++ b/R/req-promise.R
@@ -69,6 +69,12 @@ req_perform_promise <- function(req,
   check_installed(c("promises", "later"))
   check_string(path, allow_null = TRUE)
 
+  if (missing(pool)) {
+    if (!identical(later::current_loop(), later::global_loop())) {
+      cli::cli_abort("When using {.code req_perform_promise()} within {.code later::with_temp_loop()}, {.arg x} must be provided.")
+    }
+  }
+
   promises::promise(
     function(resolve, reject) {
       perf <- PerformancePromise$new(
--- a/R/req-promise.R
+++ b/R/req-promise.R
@@ -75,7 +75,10 @@ req_perform_promise <- function(req,
 
   if (missing(pool)) {
     if (!identical(later::current_loop(), later::global_loop())) {
-      cli::cli_abort("When using {.code req_perform_promise()} within {.code later::with_temp_loop()}, {.arg pool} must be provided.")
+      cli::cli_abort(c(
+        "Must supply {.arg pool} when calling {.code later::with_temp_loop()}",
+        i = "Do you need {.code pool = curl::new_pool()}?"
+      ))
     }
   }
 
--- a/R/req-promise.R
+++ b/R/req-promise.R
@@ -71,7 +71,7 @@ req_perform_promise <- function(req,
 
   if (missing(pool)) {
     if (!identical(later::current_loop(), later::global_loop())) {
-      cli::cli_abort("When using {.code req_perform_promise()} within {.code later::with_temp_loop()}, {.arg x} must be provided.")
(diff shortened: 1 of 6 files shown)


R/req-promise.R
```diff
@@ -69,6 +73,12 @@ req_perform_promise <- function(req,
   check_installed(c("promises", "later"))
   check_string(path, allow_null = TRUE)
 
+  if (missing(pool)) {
```
hadley, line 76:
```suggestion
  if (is.null(pool)) {
```

(Since missingness isn't transitive across calls, and that makes it bit more clear what happens when pool isn't missing)

R/req-promise.R
```diff
@@ -69,6 +73,12 @@ req_perform_promise <- function(req,
...
+    if (!identical(later::current_loop(), later::global_loop())) {
+      cli::cli_abort("When using {.code req_perform_promise()} within {.code later::with_temp_loop()}, {.arg pool} must be provided.")
```
hadley, line 78: I'd write this more like:

```R
cli::cli_abort(c(
  "Must supply {.arg pool} when calling {.code later::with_temp_loop()}",
  i = "Do you need {.code pool = curl::new_pool()}?"
))
```

You shouldn't need to repeat the function name because `cli_abort()` will include it by default.

R/req-promise.R
hadley: But looking at the tests suggests that an explicit `NULL` is ok?

tests/testthat/test-req-promise.R
```diff
@@ -110,3 +110,39 @@ test_that("req_perform_promise can use non-default pool", {
...
+    # proceed until the temp loop is over
+    p2 <- req_perform_promise(request_test("/get"), pool = NULL)
+
+    # You can create an async response with explicit pool=pool, and it can
+    # proceed as long as that pool was first used inside of the temp loop
+    p3 <- req_perform_promise(request_test("/get"), pool = curl::new_pool())
+
+    # You can't create an async response in the temp loop without explicitly
+    # specifying a pool
+    expect_error(p4 <- req_perform_promise(request_test("/get")))
```
hadley, line 134: I'd use `expect_snapshot(error = TRUE)` here so that you can additionally verify that the error message is informative to a human.

tests/testthat/test-req-promise.R
```diff
@@ -110,3 +110,39 @@ test_that("req_perform_promise can use non-default pool", {
...
+
+    # This works fine inside the temp loop, because its pool was first used
+    # inside
+    expect_equal(resp_status(extract_promise(p3, timeout = 1)), 200)
+  })
+
+  # These work fine now that we're back outside the temp loop
+  expect_equal(resp_status(extract_promise(p1, timeout = 1)), 200)
+  expect_equal(resp_status(extract_promise(p2, timeout = 1)), 200)
+})
```
hadley, line 148: Missing nl at end of file.

hadley (commented):
Looking great! I just commented on a few small style things.

R/req-promise.R
shikokuchuo: Why it's `missing()` is that `NULL` is the only way to specify the default pool for `curl`. `curl:::multi_default()` which returns the default pool is an internal function.

From the discussion and @jcheng5's comment here: https://github.com/shikokuchuo/httr2/pull/1#issuecomment-2466824678, I think it'd be ok if that missingness turned into `NULL` and was accepted. It was meant more as a guard against mistakes rather than a failsafe. We do also say in the docs that a new pool should be created for private loops.

R/req-promise.R
shikokuchuo: Nice, updated in [fa217a0](https://github.com/r-lib/httr2/pull/580/commits/fa217a04382685698dd1e26b23515c1e90f9166a) liking the change!

tests/testthat/test-req-promise.R
shikokuchuo: Thanks. Updated in [08b0ed5](https://github.com/r-lib/httr2/pull/580/commits/08b0ed5956e6b4c9b36e6967aea00d0eb947b222).

tests/testthat/test-req-promise.R
shikokuchuo: Also updated in [08b0ed5](https://github.com/r-lib/httr2/pull/580/commits/08b0ed5956e6b4c9b36e6967aea00d0eb947b222).

R/req-promise.R
hadley: Or if it's `NULL`, should we just automatically create a new pool instead of using the curl default pool?

shikokuchuo: I thought about this, but
(i) I'm not sure you'd want `NULL` to mean different things in different contexts, even if documented, and
(ii) it would work if you only did one request in a private loop, otherwise it would inefficiently create a new pool for each request.

The pattern we want users to be using should be something like:
```r
later::with_temp_loop(
  pool <- curl::new_pool()
  r1 <- req_perform_promise(pool = pool)
  r2 <- req_perform_promise(pool = pool)
)
```

hadley: Oh gotcha.

hadley: Let's leave it as is then

PR #589, opened 2024-11-26
Bump later requirement

shikokuchuo:
Uses the newly-released version of `later`.

I noticed that Joe moved it up to Imports at `elmer` partly as 'suggests' versions aren't enforced.

I haven't done that here, but you may wish to consider.


hadley:
I think it's important for elmer to have it in imports, since you need it for much of elmer's functionality, but it's fine in suggests here, since most httr2 users won't need it.

PR #592, opened 2024-12-02
Add missing "should" to roxygen docs

mthomas-ketchbrook:
Thanks so much for this package!  It's a huge lifesaver for a lot of my team's work.

I found what I think was a missing (and important) *"should"* keyword in the documentation for the `req_auth_basic()` function.  When adding this word to the roxygen param, I kept the line width to 80 characters.

Let me know if you'd like me to add another commit that also updates the associated `.Rd` file, or if that is part of the automated build process.

### DIFF
--- a/R/req-auth.R
+++ b/R/req-auth.R
@@ -5,8 +5,8 @@
 #'
 #' @inheritParams req_perform
 #' @param username User name.
-#' @param password Password. You avoid entering the password directly when
-#'   calling this function as it will be captured by `.Rhistory`. Instead,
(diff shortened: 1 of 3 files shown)


hadley:
@mthomas-ketchbrook in general, it's slightly more convenient if you include the generated `.Rd` file, but it's also very easy for me to add after the fact.

Thanks for fixing this!

mthomas-ketchbrook:
@hadley I will make sure to do that going forward -- thank you!

hadley:
Thank you!

PR #595, opened 2024-12-12
pass options from req_perform_iterative to req_perform

guslipkin:
This allows users to pass the `verbosity`, `mock`, and `error_call` options to `req_perform` when using `req_perform_iterative`


hadley:
I'm not sure what your end goal is here, but it doesn't look like your approach is working. I'd suggest filing an issue so we can discuss how to best implement it.

PR #596, opened 2024-12-12
Add a dedicated error class for OAuth response parsing errors

atheriel:
Not all OAuth providers in the wild return spec-compliant responses in the case of an error. A prominent example is [Facebook][0], but I've seen issues with Snowflake OAuth's token exchange flow, too. Posit Connect is not well-behaved here, either.

I've also seen weird issues in the past like authentication failues resulting in redirects to human-readable HTML pages rather than proper OAuth2 error response JSON.

All this to say: I think it can be really useful to have access to the original response object when trying to diagnose an issue.

So this commit adds a dedicated `httr2_oauth_parse` error class and embeds the response object in it. This allows consumers to write code that handles these errors with a lot more context:

```r
tryCatch(
  some_oauth_using_code(),
  httr2_oauth_parse = function(cnd) {
    cli::cli_abort(
      c(
        "Unexpected failure during an OAuth flow.",
        i = "Status code: {resp_status(cnd$resp)}",
        i = "Response content type: {resp_content_type(cnd$resp)}"
      ),
      parent = cnd
    )
  }
)
```

Unit tests are included.

[0]: https://pilcrowonpaper.com/blog/dear-oauth-providers/

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # httr2 (development version)
 
+* Errors thrown during the parsing of an OAuth response now have a dedicated
+  `httr2_oauth_parse` error class that includes the original response object
+  (@atheriel, #596).
+
 # httr2 1.0.7
 
 * `req_perform_promise()` upgraded to use event-driven async based on waiting efficiently on curl socket activity (#579).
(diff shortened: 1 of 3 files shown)


atheriel:
Right now these errors look more like the following:

```
! Failed to parse response from `client$token_url` OAuth url.
* Did not contain `access_token`, `device_code`, or `error` field.
```

or

```
! Unexpected content type "text/plain".
* Expecting type "application/json" or suffix "json".
```

hadley:
Thanks!

jennybc:
> ... redirects to human-readable HTML pages rather than proper OAuth2 error response JSON

*cough, google*

Issue #597, opened 2024-12-17
[Docs]: Need to update roxygen `@param` tag for the `after` arg of `req_retry`


jimbrig:
I noticed the [source code for req_retry](https://github.com/r-lib/httr2/blob/main/R/req-retries.R) is updated and now the `after` argument's callback function wants either `NA` or a number instead of the previous `NULL`. 

To avoid confusion, the roxygen2 `@param` docs for the `after` argument should be updated to reflect this change:

Reference:

- [Line #49 of req-retries.R](https://github.com/r-lib/httr2/blob/0ea4ec9eaed07043261cc40e364681c4a4a6af6d/R/req-retries.R#L49) 
- [Line #135 of req-retries.R](https://github.com/r-lib/httr2/blob/0ea4ec9eaed07043261cc40e364681c4a4a6af6d/R/req-retries.R#L135)

Error received:

```R
Error in `httr2::req_perform()`:
! The `after` callback to `req_retry()` must return a single number or NA, not `NULL`.
Run `rlang::last_trace()` to see where the error occurred.
```

Current [@param tag for the after argument of req_retry](https://github.com/r-lib/httr2/blob/0ea4ec9eaed07043261cc40e364681c4a4a6af6d/R/req-retries.R#L49):

```R
#' @param after A function that takes a single argument (the response) and
#'   returns either a number of seconds to wait or `NULL`, which indicates
#'   that a precise wait time is not available that the `backoff` strategy
#'   should be used instead..
```

should now be:

```R
#' @param after A function that takes a single argument (the response) and
#'   returns either a number of seconds to wait or `NA`, which indicates
#'   that a precise wait time is not available that the `backoff` strategy
#'   should be used instead.
```

(changing `NULL` to `NA`)

jimbrig:
made a quick PR: https://github.com/r-lib/httr2/pull/598

not sure if you want to keep like this or maybe adjust the function to be numeric or NA or NULL?

regardless, appreciate the amazing job with this library! much smoother developer experience than the original httr library!

PR #598, opened 2024-12-17
[Bug/Docs]: Update roxygen2 @param tag for `after` argument of `req_retry` (#597)

jimbrig:
This PR corresponds to Issue #597 and simply changed a line of roxygen2 and re-documented.


hadley:
Thanks!

PR #608, opened 2024-12-23
Implement `req_headers_redacted()`

### LINKED ISSUE #561: `req_add_secret_headers()`
That automatically redact.

hadley:
Which I think is a friendlier interface than the `.redact` argument to `req_headers()`. Fixes #561


hadley:
@mgirlich FYI

PR #609, opened 2024-12-23
Control how spaces are handed in query strings

### LINKED ISSUE #432: Option to use `+` in `req_url_query()`
For the Nominatim API[1] I need to submit my request using plus signs (+) instead of %20 (spaces). I looked at all parameters of `req_url_query()`, but couldn't find how to tweak this. Is this not yet implemented?

By the way, no rush, I have used the `httr` package in the meantime...

[1] Sample API call: https://nominatim.openstreetmap.org/search?q=135+pilkington+avenue,+birmingham&format=xml&polygon_kml=1&addressdetails=1

Thanks,
Tim

hadley:
Fixes #432


R/req-url.R
```diff
@@ -60,12 +60,17 @@ req_url <- function(req, url) {
 #'   If none of these options work for your needs, you can instead supply a
 #'   function that takes a character vector of argument values and returns a
 #'   a single string.
+#' @param .space How should spaces in query params be escaped? The default,
+#'   "percent", uses standard percent encoding (i.e. `%20%`), but you can opt-in
```
jonthegeek, line 64:
```suggestion
#'   "percent", uses standard percent encoding (i.e. `%20`), but you can opt-in
```

jonthegeek (commented):
1 typo (unless I'm missing reasoning behind the second %). This will definitely be useful! I can't remember which api it was for, but I definitely had to implement this in front of `req_url_query()` before!

hadley:
Thanks!

PR #610, opened 2024-12-23
Allow relative urls when parsing

### LINKED ISSUE #449: Need absolute URL helper
When scraping a site that uses relative urls (which is very common)

hadley:
Fixes #449


hadley:
Should also add:

```R
#' # Use a relative url
#' req <- request("http://example.com/a/b/c")
#' req |> req_url_relative("..")


#' @export
#' @rdname req_url
req_url_relative <- function(req, url) {
  check_request(req)

  new_url <- url_parse(url, base_url = req$url)
  req_url(req, url_build(url))
}
```

PR #613, opened 2024-12-23
Implement circuit breaking

### LINKED ISSUE #370: Add circuit breaking for retries
https://github.com/r-lib/httr/pull/671

hadley:
Fixes #370

@atheriel I'd love your thoughts on this once you're back from break. Does this seem like a faithful translation of your [httr PR](https://github.com/r-lib/httr/pull/671/files)?

* [x] Figure out why that  current implementation slows the tests down from ~13s to ~25s


tests/testthat/test-req-retries.R
```diff
@@ -92,3 +92,33 @@ test_that("is_number_or_na implemented correctly", {
...
+  req_perform(req) %>%
+    expect_no_condition(class = "httr2_perform") %>%
+    expect_error(class = "httr2_breaker")
+
+  # Attempt on same realm errors without trying at all
+  req2 <- request_test("/status/:status", status = 200) %>%
+    req_retry()
+  req_perform(req) %>%
+    expect_no_condition(class = "httr2_perform") %>%
+    expect_error(class = "httr2_breaker")
```
atheriel, line 123: I'm a bit surprised this works. If `req_retry()` for this request uses the default threshold of `Inf`, how does it consider the breaker tripped here?

R/req-retries.R
```diff
@@ -117,6 +132,37 @@ retry_max_seconds <- function(req) {
...
+    return(invisible())
+  }
+
+  remaining <- req$policies$retry_failure_timeout - (now - triggered)
+  if (remaining <= 0) {
+    env_unbind(the$breaker, realm)
+  } else {
+    cli::cli_abort(
+      c(
+        "Too many request failures: circuit breaker triggered for realm {.str {realm}}.",
```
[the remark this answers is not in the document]
hadley: LGTM

tests/testthat/test-req-retries.R
hadley: Once the breaker has been triggered for a realm, it stays triggered regardless of the threshold in any subsequent request (as long as that request uses retries). Does that make sense?

atheriel: Oh, I see. Yes, I think that does make sense.

hadley:
Looks like slowdown is because `req_perform` tests take an extra 20s.

PR #614, opened 2024-12-24
More tools for manipulating urls

### LINKED ISSUE #425: Export query_build and query_parse.
Functions `query_build` and `query_parse` are very useful in the day to day work of web scraping.  I think it would be great if the they were exported.

hadley:
Fixes #425


jonthegeek:
How would you feel about `url_modify_path()` (or maybe `url_modify_path_append()`) along these same lines? I know you can do it with the existing tools, but something equivalent to `req_url_path_append()` but for urls themselves is handy to (for example) avoid duplicate /'s in the same sorts of situations where you'd use these other url modify tools.

hadley:
@jonthegeek hmmmm I think I'd prefer `url_modify_relative()` now that we have more relative path handling tools. That feels a bit more general to me.

jonthegeek:
> @jonthegeek hmmmm I think I'd prefer `url_modify_relative()` now that we have more relative path handling tools. That feels a bit more general to me.

That makes sense!

PR #620, opened 2024-12-24
Look for secrets in keyring too

### LINKED ISSUE #346: Mention keyring package in Wrapping APIs vignette?
More precisely, should this workflow be added to that vignette: In an API package doc, recommend the user store their API key with keyring and retrieve it with `Sys.getenv(API_KEY=keyring::key_get("API-NAME")`?

hadley:
* [ ] Update vignette

Fixes #346

@gaborcsardi when you get back from break, I'd appreciate your feedback on this. Do you think this is a reasonable way to add keyring support to httr2 so folks can use it instead of env vars? Should httr2 use its own keyring instead of the system default?

### DIFF
--- a/tests/testthat/test-curl.R
+++ b/tests/testthat/test-curl.R
@@ -134,7 +134,7 @@ test_that("can translate data", {
 
   expect_snapshot({
     curl_translate("curl http://example.com --data abcdef")
-    curl_translate("curl http://example.com --data abcdef -H Content-Type:text/plain")
+    curl_translate("curl http://example.com --data abcdef -H Content-Type:text/plain") |>
   })
 })
 
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -36,6 +36,7 @@ Suggests:
     httpuv,
     jose,
     jsonlite,
+    keyring,
     knitr,
     later (>= 1.4.0),
     paws.common,
--- a/R/secret.R
+++ b/R/secret.R
@@ -160,6 +160,13 @@ secret_has_key <- function(key) {
 }
 
 secret_get_key <- function(envvar, call = caller_env()) {
+  if (is_installed("keyring")) {
+    key <- tryCatch(keyring::key_get(envvar), error = function(e) NULL)
+    if (!is.null(key)) {
+      return(key)
+    }
+  }
+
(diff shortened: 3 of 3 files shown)


tests/testthat/test-curl.R
```diff
@@ -134,7 +134,7 @@ test_that("can translate data", {
 
   expect_snapshot({
     curl_translate("curl http://example.com --data abcdef")
-    curl_translate("curl http://example.com --data abcdef -H Content-Type:text/plain")
+    curl_translate("curl http://example.com --data abcdef -H Content-Type:text/plain") |>
```
gaborcsardi, line 137: Typo?

gaborcsardi:
## General thoughts wrt keyring

There are a couple of things to work out wrt keyring usage in general. 

One is that it might "freeze" in a headless setup, because it starts an interactive password dialog. This is how pip deals with this: https://pip.pypa.io/en/stable/topics/authentication/#here-be-dragons
We could possibly do better. E.g. we could start a subprocess to get the password from the keyring, with a 1-2 minute (?) timeout in non-interactive sessions. We should probably also use a different keyring backend on GHA, which is the place where this happens most often (at least for us). We should probably also cache the passwords in the session, at least in interactive sessions (?), because starting up a subprocess is relatively expensive. This should all go into the keyring package, probably.

Another thing is to make sure that all packages look for the same key when looking up the password for a URL. This includes keys across languages, e.g. if there is a password for PPM in the keyring, both R (pak, httr2?, etc.) and pip should find it and use it.

A third thing to work out is to make everything smooth with gitcreds and git passwords in general. This is probably up to the R package. For some packages it makes sense to look at both the git credential store and also keyring natively (pak?), for others probably not.

## About this PR

As for this PR, maybe it would make sense to make this opt-in until we figure out some of the details? 

Should httr2 message the user about getting the password from the keyring? Or maybe only for 403 responses?

I also wonder it we should give some advice to the user. E.g. if the HTTP response is a 403, and the keyring package was not installed, or there is no password for the site.

hadley:
@gaborcsardi ok, let's leave this PR sitting for now. It feels like it'll be worth spending some time on keyring this year as part of the R in production story, and we can re-look at httr2 integration once that's done.

PR #632, opened 2025-01-06
Add `verbosity` argument to `req_peform_connection()`

### LINKED ISSUE #599: Make `resp_stream_*()` respect `verbosity`
So it's easier to debug streaming connections

See https://github.com/tidyverse/elmer/pull/226 for ideas

hadley:
Fixes #599

* [x] Check that output looks good with ellmer


R/resp-stream.R
```diff
@@ -263,24 +297,27 @@ resp_boundary_pushback <- function(resp, max_size, boundary_func, include_traile
...
     print_buffer(chunk, "Received chunk")
 
-    # If we've reached the end of input, store the buffer and return NULL
     if (length(chunk) == 0) {
       if (!isIncomplete(resp$body)) {
         # We've truly reached the end of the connection; no more data is coming
-        if (include_trailer && length(buffer) > 0) {
-          return(buffer)
-        } else {
+        if (length(buffer) == 0) {
```
hadley, line 305: @jcheng5 I don't think I've changed the logic here, just tweaked it a bit so we can warn when we get an unexpected end of input.

R/resp-stream-aws.R
```diff
@@ -57,15 +67,15 @@ parse_aws_event <- function(bytes) {
 
   # headers
   headers <- list()
-  while(i <= 12 + header_length) {
+  while (i <= 12 + header_length) {
```
hadley, line 70: Sorry about the spurious style diffs; I've been experimenting with format on save.

R/resp-stream.R
```diff
@@ -263,24 +297,27 @@ resp_boundary_pushback <- function(resp, max_size, boundary_func, include_traile
...
           return(NULL)
+        } else {
+          if (include_trailer) {
+            return(buffer)
+          } else {
+            cli::cli_warn("Premature end of input; ignoring final partial chunk")
```
jcheng5, line 311: Would it make sense to log the buffer here, under certain verbosity levels?

jcheng5 (approved):

R/resp-stream.R
hadley: Do you think that's likely to be useful? i.e. going from raw bytes to understand what's gone wrong seems like it will be a challenge.

hadley: What if we just enabled `print_buffer()` when verbosity >= 3?

PR #643, opened 2025-01-14
Tests: run the multi-req tests first

gaborcsardi:
This seems to fix the test failures. 

An alternative fix would to stop the example server in the problematic `test_that()` block and start a new one. 

Yet another alternative is to increase the number of threads to (say) 50. (They are started on demand, 50 is only the maximum.)


hadley:
Thanks!

PR #644, opened 2025-01-20
perf: set check.names to false

m-muecke:
set check.names to false and remove stringsAsFactors since using R >= 4.0

### DIFF
--- a/R/req-throttle.R
+++ b/R/req-throttle.R
@@ -62,8 +62,7 @@ throttle_status <- function() {
   data.frame(
     realm = realms,
     last_request = .POSIXct(unlist(the$throttle[realms]) %||% double()),
-    row.names = NULL,
-    stringsAsFactors = FALSE
+    check.names = FALSE
   )
 }
 


R/req-throttle.R
```diff
@@ -62,8 +62,7 @@ throttle_status <- function() {
   data.frame(
     realm = realms,
     last_request = .POSIXct(unlist(the$throttle[realms]) %||% double()),
-    row.names = NULL,
```
hadley, line 65: I think we want to keep `row.names = FALSE`?

m-muecke: `row.names` is per default set to `NULL` but I can add it back

hadley: That's a lie:

``` r
rownames(data.frame(x = c(x = 1)))
#> [1] "x"
rownames(data.frame(x = c(x = 1), row.names = NULL))
#> [1] "1"
```

<sup>Created on 2025-01-21 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

m-muecke: Sorry about that, fixed now.

hadley:
Thanks!

PR #645, opened 2025-01-21
aws_v4_signature bugfix

jeffreyzuber:
Fixed bug where 'CanonicalQueryString' was used as an argument to query-build before being defined, instead of 'sorted_query'. Before, using the function req_auth_aws_v4 would result in the error 'object 'CanonicalQueryString' not found'.


hadley:
Oops 😬  Would you mind adding a test too?

jeffreyzuber:
I thought about it, but couldn't think of a way to add a test without hard-coding some sort of AWS authentication tokens.  The best that I could come up with is to test the output of aws_v4_signature with a fixed set of arguments, including the timestamp, but that wouldn't test the actual authentication with AWS. Would that be good enough?

hadley:
Yeah, I think that would be good enough, given that the previous code wouldn't even run.

jeffreyzuber:
Added the requested test of aws_v4_signature

hadley:
Thanks! I switched your test to use a snapshot test, since the values aren't so important here and just clutter up the test file.

PR #666, opened 2025-02-07
Less-zealous redirect_uri normalization

### LINKED ISSUE #646: redirect_uri vs added /
httr2 (via `curl::curl_parse_url()`, which in turn is inheriting this behavior from C) always uses a minimum of "/" for the path now. That's almost always ok, but it caused a token request to fail for me via `normalize_redirect_uri()`, when "http://127.0.0.1:8888" (a path specified for that client) was normalized to "http://127.0.0.1:8888/" (which the client did not recognize). For this particular use-case I was able to patch in a fix by using `sub("/$", "", url_build(parsed))` as the `uri`, but that feels hacky.

I can't think of a situation other than redirect_uri where this really matters. I also ran into it yesterday in a test (the req$url now always has the trailing `/` for httr2 1.1+ and previously lacked it, which caused a macos CI test to fail when it wasn't caught up with the other OSs yet), but that didn't feel like a big deal. The redirect one prevented me from refreshing a token until I patched it, though.

This *kind of* should be fixed at the curl level, but I wouldn't always consider this "broken", so I think `normalize_redirect_uri` is probably the best place.

jonthegeek:
Closes #646

I also tweaked a snapshot test that was coming up as changed on my Windows machine. (Edit: Tweak removed, see #669)

### DIFF
--- a/tests/testthat/test-curl.R
+++ b/tests/testthat/test-curl.R
@@ -211,7 +211,7 @@ test_that("can read from clipboard", {
       clipr::read_clip()
     },
     transform = function(x) {
-      grep("\\\\r", x, value = TRUE, invert = TRUE)
+      grep('"\\\\r"\\s*$', x, value = TRUE, invert = TRUE)
     }
   )
 })
--- a/tests/testthat/test-curl.R
+++ b/tests/testthat/test-curl.R
@@ -204,16 +204,11 @@ test_that("can read from clipboard", {
   rlang::local_interactive()
 
   clipr::write_clip("curl 'http://example.com' \\\n -H 'A: 1' \\\n -H 'B: 2'")
-  expect_snapshot(
-    {
-      curl_translate()
-      # also writes to clip
-      clipr::read_clip()
-    },
-    transform = function(x) {
-      grep('"\\\\r"\\s*$', x, value = TRUE, invert = TRUE)
-    }
-  )
+  expect_snapshot({
+    curl_translate()
+    # also writes to clip
+    clipr::read_clip()
+  })
 })
 
 test_that("encode_string2() produces simple strings", {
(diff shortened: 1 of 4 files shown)


tests/testthat/test-curl.R
```diff
@@ -204,11 +204,16 @@ test_that("can read from clipboard", {
...
-  })
+  expect_snapshot(
+    {
+      curl_translate()
+      # also writes to clip
+      clipr::read_clip()
+    },
+    transform = function(x) {
+      grep("\\\\r", x, value = TRUE, invert = TRUE)
+    }
```
hadley, line 215: What's this for? Because I have had some sense that testthat was missing something, but the last time I looked into it, I couldn't find a problem: https://github.com/r-lib/testthat/pull/2004

jonthegeek: It's specifically that the *clipboard* contains a `\r` character. I'm pushing a more precise transform that I realized while working out an explanation here.

jonthegeek: I can't decide how I "wish" it worked. Maybe `clipr::read_clip()` should scrub `\r` line endings? Or maybe it's fine as-is and that might cause other issues?

hadley: Right, but `expect_snapshot()` is supposed to strip all `\r`. So why does it matter?

jonthegeek: I'd have to dig in and walk through to grok it for sure, but it feels like this is a case where the vector literally includes the "\r" character, not that the "\r" is added on a file that's read or that sort of thing. I *think* I'd be confused if I had a vector with a carriage return character and testthat silently stripped it. I also feel like there's no REAL difference there, though, so I'm not certain that it's right for it to make it through.

jonthegeek: And I guess to clarify... It wasn't stripping the \r, so I had to add this for tests to pass. I made sure everything was updated in case it was that.

hadley: Lets explore this in a separate PR, because I think this is a testthat bug.

jonthegeek: Fair enough, removing this change.

jonthegeek: See #669.

PR #676, opened 2025-02-10
Explicitly expose cache object in `req_auth_sign()`

hadley:
@atheriel can you please take a look at this? I've implemented with the goal of supporting OAuth re-authentication for `req_perform_parallel()`, but I think it generally decouples concerns a bit more clearly than previously, and it might help with some of the auth stuff that you've been experimenting with.


atheriel (approved):
This does make sense to me.

PR #684, opened 2025-02-13
Memoise default user agent

hadley:


atheriel:
It's pretty common for other packages to have "expensive" user agents like this, too. Is it possible we could make it easier for them to memoise, too?

hadley:
Yeah, but (a) it's not really that expensive (I'm only memoising here out of an abundance of caution) and (b) it's not that hard to memoise yourself.

PR #697, opened 2025-02-20
Avoid partial matching of "refresh_token"

### LINKED ISSUE #696: Refresh tokens vs annoyingly named oauth responses
Google oauth now returns `refresh_token_expires_in` as one of the fields in its response. It looks like this change happened roughly yesterday morning (the warnings first appeared in my log at 5:09am CST, but it hadn't ran since 5:14pm the day before). Because of that, [this line](https://github.com/r-lib/httr2/blob/528846a8a7f871ef8538e2272f590f9a38346679/R/oauth-token.R#L83) in `token_refresh()` now results in a warning (which really should be an error):

```r
out$refresh_token <- out$refresh_token %||% refresh_token
# Warning message:
# In out$refresh_token :
#  partial match of 'refresh_token' to 'refresh_token_expires_in'
```

Since `out` has `refresh_token_expires_in` (and not `refresh_token`), adding `refresh_token` doesn't work as expected.

I'm opening a PR momentarily to fix this, I hope.

jonthegeek:
See #696 for the full description of why this is an issue here, but the same "editing a list that we don't fully own" issue could be important in a lot of other places in this package.

A global search for `\w+\$\w+\s*<-` finds a ton of results. Is there a way to make this safer globally?

Closes #696.

### DIFF
--- a/R/oauth-token.R
+++ b/R/oauth-token.R
@@ -80,6 +80,6 @@ token_refresh <- function(client, refresh_token, scope = NULL, token_params = li
     scope = scope,
     !!!token_params
   )
-  out[["refresh_token"]] <- out[["refresh_token"]] %||% refresh_token
+  out$refresh_token <- out[["refresh_token"]] %||% refresh_token
   out
 }
--- a/R/oauth-token.R
+++ b/R/oauth-token.R
@@ -80,6 +80,6 @@ token_refresh <- function(client, refresh_token, scope = NULL, token_params = li
     scope = scope,
     !!!token_params
   )
-  out$refresh_token <- out$refresh_token %||% refresh_token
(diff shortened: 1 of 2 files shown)


jonthegeek:
OH! It's specifically this construct that's an issue:
```r
x$y <- x$y %||% "new value"
```

That's better, but really unfortunate, because I suspect I have that in a lot of places in my own code...

This regex only gets [one other hit](https://github.com/r-lib/httr2/blob/528846a8a7f871ef8538e2272f590f9a38346679/R/utils.R#L254): `(\w+\$\w+)\s*<-\s*\1\s*%\|\|%`
```
args$name <- args$name %||% name
```

I'm about to push a fix for that.

hadley:
Thanks!

PR #715, opened 2025-04-04
Update `with_mocked_responses` to use new name in docs

jansim:
The docs for `with_mocked_responses` still use `with_mock` in the examples / return values, which had me confused for a moment when using these today.

I think it makes sense to use the non-deprecated name instead? This PR replaces `with_mock` with `with_mocked_responses` in the roxygen comments.


hadley:
Oops, good catch, thanks!

PR #723, opened 2025-05-03
re-calculate n lengths after exploding

### LINKED ISSUE #719: multi_dots doesn't re-calculate n lengths after exploding
Ran into an issue when using req_url_query(.multi = "explode") with a mix of multi-length and NULL params. It looks like n isn't re-calculated after exploding in multi_dots() on [L66](https://github.com/r-lib/httr2/blob/main/R/utils-multi.R#L66) which causes n to be shorter than dots creating a select/recycling issue.

```r
expect_equal(
  multi_dots(a = 1:2, b = NULL, .multi = "explode"),
  list(a = I("1"), a = I("2"), b = NULL)
)

expect_equal(
  multi_dots(a = 1:3, b = NULL, .multi = "explode"),
  list(a = I("1"), a = I("2"), a = I("3"), b = NULL)
)
```

Since it's using explode, then I would think that n would always have a value of 0 or 1 removing the need of the n > 1 handling for explode. I think it could just have the following for the `.multi == "explode"` section?

```r
dots <- explode(dots)
n <- lengths(dots)
```

Kevanness:
Closes #719

- Add a test for mixed length query params
- Replace handling of n > 1 in .explode with re-calculating lengths.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httr2 (development version)
 
+* `multi_dots()` now re-calculates n lengths when using `.multi = "explode"` to avoid select/recycling issues (@Kevanness, #719).
+
 # httr2 1.1.2
 
 * `req_headers()` more carefully checks its input types (#707).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # httr2 (development version)
 
-* `multi_dots()` now re-calculates n lengths when using `.multi = "explode"` to avoid select/recycling issues (@Kevanness, #719).
+* `req_url_query()` now re-calculates n lengths when using `.multi = "explode"` to avoid select/recycling issues (@Kevanness, #719).
 
 # httr2 1.1.2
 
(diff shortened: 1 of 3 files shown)


hadley:
Looks good! Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

Kevanness:
Added the news bullet, wasn't really sure if it was needed since it wasn't a user-facing change.

hadley:
Isn't it a user facing problem? I thought it created an incorrect output?

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # httr2 (development version)
 
+* `multi_dots()` now re-calculates n lengths when using `.multi = "explode"` to avoid select/recycling issues (@Kevanness, #719).
```
hadley, line 3:
```suggestion
* `req_url_query()` now re-calculates n lengths when using `.multi = "explode"` to avoid select/recycling issues (@Kevanness, #719).
```

hadley:
Thank you!

Issue #718, opened 2025-04-09
Programmatically get request body data


sckott:
via https://github.com/ropensci/vcr/issues/276 discussion

There may be other use cases, but the use case for this right now is in `vcr` we want to record in a vcr cassette the same information that is sent in a real HTTP request for the request body. 

There is `httr2:::req_body_get()` but it is not exported for one, for two it doesn't cover all body types, and three we'd need a way to gather the data that is sent in a real request for the request body. 

One can run `req_dry_run` to see what is sent but AFAICT the request body data isn't available from that call anyway. 

Ideally for the `vcr` use case there'd be a way to get the request body data as it would be sent in a real http request BEFORE making the request, because although we don't record the cassette until after the real request is made on first recording, on subsequent recordings of the matching request we'd need a way to determine the request body data WITHOUT having to make a real HTTP request (assuming that is, the user is matching on request body)

hadley:
Also need to export `req_method_get` so you don't need `webmockr:::req_method_get_w()`.

Or maybe we could export `req_prepare()` which does `req_method_apply()` and `req_body_apply()`

PR #738, opened 2025-06-13
Clarify body types

hadley:
General refactoring of code and tests in prepation for #718. It should now be much clearer what sorts of body types are permissible. Includes a few small improvements to error messages which I noticed while adding additional tests to get to 100% test coverage.


sckott:
@hadley  Thanks for the heads up. I just pulled down main and indeed breaks `vcr` tests. i'll work on fixing those issues. It might break `webmockr` too - still looking into that.

PR #741, opened 2025-06-16
Use curl to modify/construct urls

### LINKED ISSUE #732: Use `curl_modify_url()` instead of home grown

hadley:
Note that `req_url_path_append()` now only actually affects the path, not other components of the URL like the query params.

To do:

* [ ] revdepcheck

Fixes #732


hadley:
The revdeps seems to be mostly something to do with httptest2, so I should figure that out first. But I only see one very obvious failure related to this change, and that's atrrr.

hadley:
And a couple of failures due to the case of the encoded characters changing.

hadley:
Oooh actually the httptest2 failures might be because of that case change.

PR #742, opened 2025-06-17
Don't serialize redacted headers

### LINKED ISSUE #721: Make redacted field more robust
Maybe encrypt/decrypt using httr2 local key? That would prevent automated tools from finding it.

hadley:
This required quite a lot of refactoring to clarify what's actually going on. It should now be clearer that there is a redacted sentinel value which is used to nicely format, print, and str redacted values that might be exposed to the user (I don't fully recall why I added this but I suspect it's because we only want to apply cli formatting at the exact time when the value is displayed). This code now lives in `utils-redacted.R`.

I also clarified the type of the headers object — the componets can either be an atomic vector or a weakref. And since redacted components now have their own type, we no longer need the `redact` attribute.

Fixes #721


shikokuchuo:
Practically there's no way to access the 'value' from the 'key' if the weakref itself is no longer around, hence it's probably fine that you tie the lifetime of 'value' to the package `the` environment.

However, if you want to ensure the lifetime of 'value' is really tied to a single request, then you could instead pass the request `req` as an argument to `as_header()` and consuming functions, and you'd then be able to use `req$state` as the 'key'.

hadley:
I decided to make the lifespan a bit more explicit and I do think it makes it clearer.

PR #744, opened 2025-06-18
Export `req_get_method()` and `req_get_body()`

### LINKED ISSUE #718: Programmatically get request body data

hadley:
Formerly known as `req_method_get()` and `req_body_get()`, but I think it's better to put exported accessors behind a common prefix.

Fixes #718

@sckott does this look good to you?


sckott:
For body, look like we can use `req_get_body()` here to replace `vcr:::httr2_body()` https://github.com/ropensci/vcr/blob/main/R/request_handler-httr2.R#L15 

And for method, replace the two instances of `webmockr:::req_method_get_w` in `request_handler-httr2.R` with `req_get_method()`

I made those two changes locally in `vcr` and one failure in particular caught my eye:

This block fails https://github.com/ropensci/vcr/blob/main/tests/testthat/test-request_handler-httr2.R#L221-L236 

```r
[Cassette: test] Inserting 'test.yml' (with 1 interactions)
[Cassette: test]   recording: FALSE
[Cassette: test] Handling request: POST http://127.0.0.1:57919/post --------------------------PWp7LJZVTwCltMve42T14D
Content-Disposition: form-data
[Cassette: test]   Looking for existing requests using body
[Cassette: test]     Request 1: NO MATCH
[Cassette: test]       lines(matching$body) vs lines(recorded$body)
[Cassette: test]       - "--------------------------PWp7LJZVTwCltMve42T14D\r"
[Cassette: test]       + "--------------------------9t6nwKJJSFOWermXv95b6Q\r"
[Cassette: test]         "Content-Disposition: form-data; name=\"a\"\r"
[Cassette: test]         "\r"
[Cassette: test]         "x\r"
[Cassette: test]       - "--------------------------PWp7LJZVTwCltMve42T14D\r"
[Cassette: test]       + "--------------------------9t6nwKJJSFOWermXv95b6Q\r"
[Cassette: test]         "Content-Disposition: form-data; name=\"b\"\r"
[Cassette: test]         "\r"
[Cassette: test]         "y\r"
[Cassette: test]       - "--------------------------PWp7LJZVTwCltMve42T14D--\r"
[Cassette: test]       + "--------------------------9t6nwKJJSFOWermXv95b6Q--\r"
[Cassette: test]         ""
[Cassette: test]   No matching requests
[Cassette: test] Ejecting
```

I'm guessing boundary delimiters need to be different for every request even if the data is the same? Perhaps we can mock that somehow in vcr?

hadley:
@sckott yeah, the boundaries are random. You can see how I deal with it here: https://github.com/r-lib/httr2/pull/744/files#diff-2219be50cb219172a9e6a0564807186e38b9ff2307ed0bc5f5b72d45c45585deR197

PR #745, opened 2025-06-18
Bump requirement to R 4.1

hadley:
And use base pipe and lambda.

@gaborcsardi did I miss anything when removing our old examples hack?


gaborcsardi:
> @gaborcsardi did I miss anything when removing our old examples hack?

AFAICT you got everything.

PR #746, opened 2025-06-18
Add interrupts escape hatch

### LINKED ISSUE #720: Parallel interrupts problem
```R
repeat {
  req_parallel_requests()
}
```

is very hard to escape from. And you can't solve it by wrapping in another `tryCatch()`:

```R
tryCatch(
  repeat {
    req_parallel_requests()
  },
  interrupt = function(cnd) stop("!")
}
```

because the interrupt is captured and doesn't bubble up. 

Two possible ideas:

* Record the time between interrupts and if it's less than a threshold (suggesting that the user has pressed Ctrl + C twice in a row), resignal the interrupt.
* Resignal something that an enclosing scope could capture

(or possibly both)

hadley:
I can't see any way to test this, but I experimented interactively with this code:

```R
library(httr2)
request_base <- request(example_url())
reqs <- rep(list(request_base |> req_url_path("/delay/0.1")), 100)
repeat(req_perform_parallel(reqs))
```

Fixes #720

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # httr2 (development version)
 
+* Functions that capture interrutps (like `req_perform_parallel()` and friends) are now easier to escape if they're called inside a loop: you can press Ctrl + C twice to guarantee an exit (#1810).
 * New `last_request_json()` and `last_response_json()` to conveniently see JSON bodies (#734).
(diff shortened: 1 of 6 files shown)


hadley:
Thanks for the review!

Issue #726, opened 2025-05-19
req_perform_parallel() string parameter not working since version 1.1.1


Eli-Berkow:
With version 1.1.0 and prior, the following example works with a string for the progress parameter:

```
library(shiny)
library(httr2)

ui <- fluidPage(
    htmlOutput('resps')
)

server <- function(input, output, session) {
    output$resps <- renderUI({
        request_base <- request(example_url())
        reqs <- list(
            request_base |> req_url_path("/delay/1"),
            request_base |> req_url_path("/delay/1")
        )
        parallel_progress=req_perform_parallel(reqs, progress = 'Running')
        HTML('done')
        })
}

shinyApp(ui, server)
```

![Image](https://github.com/user-attachments/assets/ba48d08f-7f2a-49c4-a24c-8367b6890b20)

Replacing the string 'Running' with TRUE gives the below default:

![Image](https://github.com/user-attachments/assets/f4f62c5b-a097-4d20-9018-a9ff3ec813e9)

From version 1.1.1 there is no progress bar at all for strings and the following shows for the default TRUE option:

![Image](https://github.com/user-attachments/assets/812cc59c-98ae-4066-9440-27a8658eb942)

Has there been a change in the way that parameter is used?
Thanks in advance

PR #755, opened 2025-06-23
Make `req_perform_parallel` take custom formatting options (#726)

fwimp:
`req_perform_parallel()` does not correctly handle progress bar formatting arguments as of v1.1.1 (see #726)

This PR transitions `req_perform_parallel()` to the progress bar creation method used by other `req_perform_*()` functions (specifically using the `create_progress_bar()` function from utils.R)

Extra data required for the default progress bar style is passed through to the bar using the `extras` functionality of `cli`.

When providing a string as the `progress` argument, the bar uses the default `[working] (1 + 4) -> 5 -> 5⁠` format, but prepends the user-provided string:

`progress="MyDownload"` leads to a bar format of `MyDownload[working] (1 + 4) -> 5 -> 5⁠`

### DIFF
--- a/R/req-perform-parallel.R
+++ b/R/req-perform-parallel.R
@@ -154,16 +154,46 @@ RequestQueue <- R6::R6Class(
     ) {
       n <- length(reqs)
 
+      barname <- ""
+
+      if (is.character(progress)){
+        barname <- progress
+        progress <- TRUE
+      }
+
       if (isTRUE(progress)) {
-        self$progress <- cli::cli_progress_bar(
+        self$progress <- create_progress_bar(
           total = n,
-          format = paste0(
-            "[{self$queue_status}] ",
-            "({self$n_pending} + {self$n_retries}) -> {self$n_active} -> {self$n_complete} | ",
-            "{cli::pb_bar} {cli::pb_percent}"
-          ),
-          .envir = error_call
+          name = barname,
+          config = list(format = paste0(
+                "{cli::pb_name}[{cli::pb_extra$queue_status}] ",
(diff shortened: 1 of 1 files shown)


R/req-perform-parallel.R
```diff
@@ -154,16 +154,46 @@ RequestQueue <- R6::R6Class(
...
       n <- length(reqs)
 
+      barname <- ""
+
+      if (is.character(progress)){
+        barname <- progress
+        progress <- TRUE
+      }
+
       if (isTRUE(progress)) {
```
hadley, line 164: I think this would be a bit simpler to tackle if you added a `format` to `create_progress_bar()`. And you shouldn't need to specify `extra` because `create_progress_bar()` should be setting the environment so that is not necessary.

fwimp: Yeah that's a good point, I didn't want to change more than was strictly necessary (especially since quite a few funcs depend on `create_progress_bar()`).

And that was what I thought too, but no matter what I tried, cli was kicking out "cannot find variable `queue_status`"/"cannot find variable `self$queue_status`" etc. messages when I tried without passing through extras.

If I change the `create_progress_bar()` call's environment to `current_env()` instead of `caller_env()` then the progress bar now has access to the data to update from, but then it seems to create the progress bar one environment too deep and then loses access with a `Cannot find progress bar cli-53424-2325` when it tries to update it. 

I'm sure this is more an issue with my understanding of environment inheritance. I'll come back at it tomorrow.

fwimp: From a `browser()` call at L203:
image

fwimp: I look forward to seeing your fix! The minutia of environment inheritance still escape me a bit in all honesty.

PR #756, opened 2025-06-23
Support mocking in `req_perform_connection()`

### LINKED ISSUE #651: Bring mocking ability to other req_perform* functions?
Of the various `req_perform*` functions (thinking mostly of `req_perform_parallel` and `req_perform_connection`) I think `mock` is only available in `req_perform`. Is that wrong? 

Would you consider adding mocking ability (whether through a parameter or via a with handler like `with_mocked_responses`) to the other `req_perform*` functions? 

Use case is so that packages like vcr, webmockr, httptest2 can hook into them

hadley:
This is supported by a new `StreamingBody` R6 class. This class is exported so that you can instantiate it or extend it, as needed.

Fixes #651 

@sckott, @nealrichardson if you have a chance I'd love to get your eyes on this. The easiest way to see how this will work is to look at [the test](https://github.com/r-lib/httr2/pull/756/files#diff-0fcb5cf8946210246bedcf78d8437de900739fbda89d74a5310abc6466c9ca2cR73-R91).

Current limitations:

* There's no way for the mocking function to know if it's being called from a streaming or non-streaming context. I suspect this is not that important, since you'll either be testing `req_perform()` or `req_perform_connection()` not both.
* If you want to mock something like `resp_stream_sse()` you're basically forced to capture the `resp_body_raw()` and then replay that. You can certainly imagine wanting to more support to represent the stream in json.

I _think_ those are probably both fine for the initial release.


hadley:
I added an internal way to tell if the request wants a connection: `req$policies$connection`. I added this because I have a sneaking suspicion we will need it; if that's true, I'll turn that into an exported API.

hadley:
Merging, not because I think it's perfect, but I think it should advance the state of things a little bit and we can iterate in the next release.

PR #763, opened 2025-06-24
Allow interactive to be overriden in req_oauth_device.

plietar:
When running in an interactive session, the oauth device flow opens the verification URI in the user's browser. When running non-interactively, the URI is instead printed to the console.

This heuristic does not always work, as for example an interactive R session running within an SSH session will not be able to open up the browser. In those cases, one can override the behaviour by using `rlang::with_interactive` when calling `oauth_flow_device`.

With `req_oauth_device` however, overriding this behaviour is more difficult. While the request might eventually call out to `oauth_flow_device`, it will do so later from within `req_perform`, after `req_oauth_device` has returned. To change the behaviour of the oauth flow one would need to wrap the entire request processing with `with_interactive`.

This commit adds a `open_browser` parameter to the `oauth_flow_device` and `req_oauth_device` functions, allowing more precise control over the behaviour. The default is unchanged, `open_browser` will be set to `TRUE` if the session is running interactively, and `FALSE` if not.  It does make it easier to disable the functionality, even when using req_oauth_device.


hadley:
Thank you for working on this!

PR #781, opened 2025-07-17
Hack around colon escaping

hadley:


jeroen (approved):
Thanks, lgtm. I've made it slightly more cautious.

PR #788, opened 2025-07-24
Pass original URL to curl_modify_url() to keep original encoding

### LINKED ISSUE #786: htt2 >= 1.2.0 replaces '%2f' with '/' in the URL
Which is probably something that it should not do?

So e.g. this does not work any more:
```r
❯ gh::gh("/orgs/r-hub/packages/container/containers%2fatlas/versions", per_page = 5)
Error in `gh::gh()`:
! GitHub API error (404): Not Found
✖ URL not found: <https://api.github.com/orgs/r-hub/packages/container/containers/atlas/versions?per_page=5>
ℹ Read more at <https://docs.github.com/rest>
```

```r
❯ rlang::last_trace()
<error/github_error>
Error in `gh::gh()`:
! GitHub API error (404): Not Found
✖ URL not found: <https://api.github.com/orgs/r-hub/packages/container/containers/atlas/versions?per_page=5>
ℹ Read more at <https://docs.github.com/rest>
---
Backtrace:
    ▆
 1. └─gh::gh(...)
Run rlang::last_trace(drop = FALSE) to see 4 hidden frames.
```

jeroen:
This changes the logic of `modify_url()` to start with the original URL and modify it, rather than build it from scratch from the list of its components. 

This should prevent normalisation of components of the URL that we do not modify, in particular the escaped slash.

Fixes #786

### DIFF
--- a/tests/testthat/test-url.R
+++ b/tests/testthat/test-url.R
@@ -127,6 +127,12 @@ test_that("path always starts with /", {
   expect_equal(url_modify("https://x.com/abc", path = NULL), "https://x.com/")
 })
 
+test_that("escaped slashes are retained", {
+  url <- "http://x.com/a%2Fb/"
+  expect_equal(url_modify(url), url)
+  expect_equal(url_modify(url, query = list(x=1)), "http://x.com/a%2Fb/?x=1")
+})
+
 # relative url ------------------------------------------------------------
 
 test_that("can set relative urls", {
--- a/tests/testthat/test-url.R
+++ b/tests/testthat/test-url.R
@@ -127,7 +127,7 @@ test_that("path always starts with /", {
   expect_equal(url_modify("https://x.com/abc", path = NULL), "https://x.com/")
 })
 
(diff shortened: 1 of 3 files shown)


tests/testthat/test-url.R
```diff
@@ -127,6 +127,12 @@ test_that("path always starts with /", {
   expect_equal(url_modify("https://x.com/abc", path = NULL), "https://x.com/")
 })
 
+test_that("escaped slashes are retained", {
```
hadley, line 130:
```suggestion
test_that("only modifies specified components", {
```

hadley (approved):
LGTM. Want to add a news bullet and then I'll resubmit to CRAN in a couple of days?

jeroen:
> LGTM. Want to add a news bullet and then I'll resubmit to CRAN in a couple of days?

I'm not sure if it's urgent to release this. It's a really unusual case, I doubt anyone else is affected.

hadley:
Ok, I'll just do a release when it feels right then 😄

PR #800, opened 2025-08-28
Replace rappdirs by tools::R_user_dir()

Bisaloo:
I see that this change was made in other r-lib repos so I thought I'd suggest it here as well.

`tools::R_user_dir()` was introduced in R 4.0 so safe to use with the tidyverse R version support policy/


hadley:
I think this probably needs a bit more thought than in other packages, because we'd need to make sure that this directory eventually gets cleaned up.

hadley:
httr2 has `cache_disk_prune()` which is called automatically on load, and ensures that all tokens > 30 days old are deleted.

hadley:
Ok, I've implemented what I think is needed. I just need to think through if it's ok for this to go in a patch release.

hadley:
I think this is probably safe for a patch release, but I just feel weird about moving these high security files without more info, i.e. the blog post that will come with a minor release. So I'm going to hold off until httr2 does its next minor release.

hadley:
I think I will do this in the next release because:

* I discovered that `httr2::oauth_cache_clear()` never actually worked
* rlang 1.3 changes the hashing algorithm so no existing tokens will be used anyway.

PR #802, opened 2025-09-09
Use base pipe

gaborcsardi:
A couple of test files still had the magrittr pipe.


hadley:
Thanks! I now it's just as easy to fix them yourself, but if you feel like filing issues, these are good for TDD.

hadley:
@jeroen we still re-export the magrittr pipe and don't have a plan to remove that yet (since it might break revdeps/user code). That said, if we did start removing somewhere, it would be reasonable to start with httr2.

Issue #825, opened 2025-12-09
req_oauth_bearer_jwt requires claim twice


drmowinckels:
I've found that to use `req_oauth_bearer_jwt` the `jwt_claim` needs to be passed in two instances for httr2 to not error.
It needs to be

- in the function it self (else function errors with "claim missing with no default")
- in the client auth_params (else errors with "client needs jwt_claim in auth_params when using jwt_sig")

Is it really necessary to pass the claim twice, or should not the claim be passed from the client to the function (like the "key" parameter is)?

```r
claim <- jwt_claim()

httr2::req_oauth_bearer_jwt(
  claim = claim,
  client = meetupr_client(
    key = "/path/to/key.pem",
    auth = "jwt_sig",
    auth_params = list(
      claim = claim
    )
  )
)
```

PR #826, opened 2025-12-09
use jwt claim in client

drmowinckels:
I believe this should fix the problem with double claims needed for jwt req (#825).


hadley:
Thanks for taking a stab at this but I ended up going in slightly different direction in #855, probably because I didn't think about it enough before asking you to have a go. I hope I didn't waste too much of your time 😞

drmowinckels:
nah, its all good. it was fun to have a stab at it, but agree that your solution is much neater.

Issue #683, opened 2025-02-12
Eliminate `Sys.sleep()` in streaming tests


hadley:
By using nanoext to sychronise:

```R
test_that("can join lines across multiple reads", {
  speaker <- nanonext::socket("req", dial = "ipc://httr2-test")
  withr::defer(unlink("httr2-test"))
  
  req <- local_app_request(function(req, res) {
    listener <- nanonext::socket("rep", listen = "ipc://httr2-test")

    res$send_chunk("This is a ")
    nanonext::recv(listener, block = TRUE)
    res$send_chunk("complete sentence.\n")
  })

  # Non-blocking returns NULL until data is ready
  resp1 <- req_perform_connection(req, blocking = FALSE)
  withr::defer(close(resp1))

  out <- resp_stream_lines(resp1)
  expect_equal(out, character())
  expect_equal(resp1$cache$push_back, charToRaw("This is a "))

  nanonext::send(speaker, "next", block = TRUE)
  out <- resp_stream_lines(resp1)
  expect_equal(out, "This is a complete sentence.")
})
```

Would need to figure out how to make a bit easier to reuse so we don't need so much boilerplate in each function.

shikokuchuo:
To synchronise 2 processes, the typical pattern is one req/rep cycle: send - recv - ack(nowledgement).

So in the below I've added an ack by the app once it's sent the next chunk, and the test waits for that before trying to read it.

Some other small cleanups to make it CRAN-ready:
- Skip if nanonext not installed.
- Use `tempfile()`. Technically the unlink is unnecessary as it's just a temporary file for unix domain sockets, but safer to keep it for CRAN "black swan" testing. On Windows this is just a name for a pipe and is not created on the filesystem.
- Use a maximum block time (again CRAN-paranoia, shouldn't be an issue when using IPC).

```r
test_that("can join lines across multiple reads", {
  skip_if_not_installed("nanonext")
  tmp <- tempfile()
  url <- sprintf("ipc://%s", tmp)
  speaker <- nanonext::socket("req", dial = url)
  withr::defer(unlink(tmp))
  
  req <- local_app_request(function(req, res) {
    listener <- nanonext::socket("rep", listen = url)
    
    res$send_chunk("This is a ")
    nanonext::recv(listener, block = 1000)
    res$send_chunk("complete sentence.\n")
    nanonext::send(listener, NULL, block = 1000)
  })
  
  # Non-blocking returns NULL until data is ready
[...]

PR #830, opened 2025-12-22
Fix synchronization in streaming tests

shikokuchuo:
Follow up to #703 in addressing #683.

The streaming tests coordinate a separate webfakes server with the client over nanonext, but those messages outrun the HTTP data: a message confirms a chunk was *sent*, not that it *arrived* or that the stream *closed*. #703's fixed 50ms delay mostly masked this but still flaked on CI (usually macOS/Windows).

The tests now wait for the actual state instead of a timer:

- `wait_for_http_data()` — until the next chunk has arrived
- `wait_for_complete()` — until the connection has closed, before reads that need the stream finished

The nanonext message just triggers the next chunk.
Also drops curl's non-deterministic `shutting down connection` line from the `verbosity = 3` snapshot.

### DIFF
--- a/R/req-verbose.R
+++ b/R/req-verbose.R
@@ -127,6 +127,7 @@ req_verbose_test <- function(req) {
 transform_verbose_response <- function(lines) {
   lines <- gsub(example_url(), "<webfakes>/", lines, fixed = TRUE)
   lines <- lines[!grepl("^<- (Date|ETag|Content-Length):", lines)]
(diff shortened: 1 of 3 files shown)


shikokuchuo:
These tests have been re-run 5 times, which was enough to trigger an error previously.

@hadley this should finally rid us of the test flakiness at httr2! (I set Claude on this today and it seems to have succeeded where it failed before)

hadley:
Thanks!!

PR #832, opened 2026-01-14
Fix `last_response_json()` to work with 'application/problem+json' 

### LINKED ISSUE #782: Should `last_response_json()` work with 'application/problem+json' content type?
Currently  `last_response_json()`  errors with  `application/problem+json` response:

```r
last_response_json()
# Error in `last_response_json()`:
# ! Last response doesn't have a JSON body.
```
But this works:
```r
last_response() |> resp_body_json()
```

A quick look at the source shows that:
- `resp_body_json()` uses `resp_check_content_type` and 
-  `last_response_json()` checks `resp_content_type(resp)` if it is "application/json"

cgiachalis:
closes #782

```r
library(plumber2)
library(httr2)

api <-  api() |>
  api_get(
    "/health",
    function() {
      plumber2::abort_bad_request("bad request")
    }
  )

mirai::daemons(1)

# run api in a separate process & do not block the current process
mi <- mirai::mirai({
  api$ignite(block = TRUE, showcase = FALSE)
}, api = api)


request("http://127.0.0.1:8080/health") |>  req_perform()

# Error in `req_perform()`:
# ! HTTP 400 Bad Request.
```

```r
 last_response_json()
{
  "type": "https://datatracker.ietf.org/doc/html/rfc9110#section-15.5.1",
  "title": "Bad Request",
  "status": 400,
  "detail": "bad request"
}

```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,5 @@
 # httr2 (development version)
 
-## Minor improvements and bug fixes
-
 * `last_response_json()` now works with content-types that end with `+json`, 
 e.g., `application/problem+json` (@cgiachalis, #782).
 


NEWS.md
```diff
@@ -1,5 +1,10 @@
 # httr2 (development version)
 
+## Minor improvements and bug fixes
```
hadley: Just include the bullet please, no headings (we add those, if necessary, as part of the release process).

PR #836, opened 2026-03-03
req_body_form() sets req to an empty list

### LINKED ISSUE #835: Request with empty form data fails
As of recent versions (perhaps 1.2.0), it is no longer possible to perform a request that has empty form data:

``` r
library(httr2)
request("http://example.com") |> req_body_form() |> req_perform()
#> Error in `url_query_build()`:
#> ! `query` must be a named list, not `NULL`.
```

<sup>Created on 2026-03-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

While this may seem like a strange thing to do, I noticed it because an API client library I wrote generated such requests depending on parameters. If I edit the request body to insert an empty list instead of null then all proceeds as expected:

``` r
library(httr2)
req <- request("http://github.com") |> req_body_form()
req$body$data <- list()
req_perform(req)
#> <httr2_response>
#> GET https://github.com/
#> Status: 200 OK
#> Content-Type: text/html
#> Body: In memory (561725 bytes)
```

<sup>Created on 2026-03-03 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

The validation code expects that `req$body$data` should be non-null, so it seems like a bug that `req_body_form()` with no arguments should create it as NULL when an empty list would avoid all this.

arcresu:
Ensures that when `req$body$data` is `NULL`, an empty list is used instead. Note that this is exactly what is done already a few lines above for `req_body_json_modify()` which is conceptually doing the same thing.

Fixes #835.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 * `oauth_flow_auth_code()` now correctly uses the same redirect URI for both authorization and token requests when using the default localhost redirect URL (@pedrobtz, #829).
 * `last_response_json()` now works with content-types that end with `+json`, 
 e.g., `application/problem+json` (@cgiachalis, #782).
+* `req_body_form()` now creates a valid empty request body when no parameters
+  are provided (@arcresu, #836).
 
 # httr2 1.2.2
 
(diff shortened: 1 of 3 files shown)


hadley:
Thanks! Can you please add a bullet to the top of `NEWS.md`? It should briefly [describe the change](http://style.tidyverse.org/news.html) and end with `(@yourname, #issuenumber)`.

arcresu:
Done!

hadley:
Thank you!

PR #845, opened 2026-06-15
Add `oauth_server_metadata()` for OAuth/OIDC discovery

shikokuchuo:
Closes #844.

Adds `oauth_server_metadata()`, which fetches and parses an OAuth 2.0 Authorization Server Metadata document ([RFC 
  8414](https://datatracker.ietf.org/doc/html/rfc8414)) or OpenID Connect Discovery document, returning the endpoints an issuer advertises.

This lets users discover values like `authorization_endpoint`, `token_endpoint`, and `device_authorization_endpoint` rather than hard-coding them:

```r
meta <- oauth_server_metadata("https://accounts.google.com")
client <- oauth_client("id", token_url = meta$token_endpoint, secret = "...")
oauth_flow_auth_code(client, auth_url = meta$authorization_endpoint)
```

PR #846, opened 2026-06-16
Let `oauth_client()` carry its server endpoints

### LINKED ISSUE #591: Reconsider app/client design based on OIDC
e.g. https://josiahparry.com/posts/2024-11-28-httr2-oidc

shikokuchuo:
Closes #591.

Builds on `oauth_server_metadata()` (#845), which returns a provider's discovered endpoints. This PR makes use of this more ergonomic by not requiring you to hand-thread those endpoints into both `oauth_client(token_url = ...)` **and** each flow's `auth_url = ...`.

### Changes

`oauth_client()` gains a single `metadata` argument. Pass the result of `oauth_server_metadata()` and the client holds all of the server's endpoints, so the flows source them automatically:

```r
metadata <- oauth_server_metadata("https://accounts.google.com")
client   <- oauth_client("id", metadata = metadata, secret = "...")
token    <- oauth_flow_auth_code(client)   # auth_url comes from the client
```

The OAuth vignette is updated to primarily use this new method.

### Design considerations

- Merge endpoints onto the client, over passing `metadata` to each flow. The client already carries `token_url` — itself a server-owned endpoint — so it was already a partial merge + matches `oauth_client()`'s own documented definition of a client (credentials + endpoints + auth mechanism).

- Store translated endpoint strings, not the raw document (client gets `token_url`, `auth_url`, `device_auth_url` resolved from the metadata fields)

- Backward compatible. `token_url` stays (now optional, defaulting from `metadata$token_endpoint`); the flows' `auth_url` becomes optional, falling back to the client, with an explicit value always winning. Existing positional calls are unaffected.

- Validation. A `metadata` object is validated whenever supplied — even alongside an explicit `token_url` — so a wrong-class object fails at the call site rather than several frames deeper.

### DIFF
--- a/R/oauth-client.R
+++ b/R/oauth-client.R
@@ -1,13 +1,15 @@
 #' Create an OAuth client
 #'
-#' An OAuth app is the combination of a client, a set of endpoints
-#' (i.e. urls where various requests should be sent), and an authentication
-#' mechanism. A client consists of at least a `client_id`, and also often
-#' a `client_secret`. You'll get these values when you create the client on
-#' the API's website.
+#' An OAuth client is the combination of a set of credentials (at least a
+#' `client_id`, and often a `client_secret` or `key`), the server endpoints it
+#' talks to, and an authentication mechanism. You'll get the credentials when
+#' you register the client on the API's website. Supply the `token_url`
+#' directly, or pass the result of [oauth_server_metadata()] as `metadata` to
+#' fill in all of the server's endpoints at once.
 #'
 #' @param id Client identifier.
-#' @param token_url Url to retrieve an access token.
+#' @param token_url Url to retrieve an access token. Not needed if `metadata`
+#'   is supplied, which sets it from the `token_endpoint`.
 #' @param secret Client secret. For most apps, this is technically confidential
 #'   so in principle you should avoid storing it in source code. However, many
 #'   APIs require it in order to provide a user friendly authentication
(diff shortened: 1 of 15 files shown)


R/oauth-client.R
```diff
@@ -83,6 +105,8 @@ oauth_client <- function(
       secret = secret,
       key = key,
       token_url = token_url,
+      auth_url = auth_url,
```
hadley, line 108: Maybe it would be cleaner to put `metadata` here, or maybe have a `urls` list? What do you think?

R/oauth-server-metadata.R
```diff
@@ -3,14 +3,17 @@
...
+#' result to [oauth_client()] as `metadata` and the endpoints are wired up for
+#' you:
 #'
 #' ```r
-#' meta <- oauth_server_metadata("https://accounts.google.com")
-#' client <- oauth_client("id", token_url = meta$token_endpoint, secret = "...")
-#' oauth_flow_auth_code(client, auth_url = meta$authorization_endpoint)
+#' metadata <- oauth_server_metadata("https://accounts.google.com")
+#' client <- oauth_client("id", metadata = metadata, secret = "...")
+#' token <- oauth_flow_auth_code(client)
```
hadley, line 16: Much nicer!

hadley (commented):
One bigger question before I do a close read.

R/oauth-client.R
shikokuchuo: It makes sense now as we're only ever extracting 3 URLs from `metadata`. I don't think it matters too much given it's not user-facing (opaque behind the S3 class). Also it means we can continue using the current print method.

vignettes/articles/oauth.Rmd
```diff
@@ -62,30 +63,67 @@ Overall this leads to a hierarchy of credentials from weakest to strongest:
...
+Most OAuth providers expose a handful of **endpoints**: URLs where httr2 sends the various OAuth requests.
+The two you'll almost always need are the authorization URL, where the user logs in and grants access, and the token URL, where httr2 exchanges that grant for an access token.
+Some providers also offer a device authorization URL, used by the device flow.
+
+Traditionally you had to dig these URLs out of the provider's documentation, but many providers now publish them in a standard discovery document (defined by [RFC 8414](https://datatracker.ietf.org/doc/html/rfc8414) and OpenID Connect Discovery).
+`oauth_server_metadata()` fetches and parses that document, so you can look up every endpoint from a single issuer URL:
+
+```{r}
+#| eval: false
+metadata <- oauth_server_metadata("https://github.com/login/oauth")
```
hadley, line 80: How do you find that url?

hadley (approved):

vignettes/articles/oauth.Rmd
shikokuchuo: There doesn't seem to be an authoritative source/database of those, so I still direct ppl to provider docs in [398b4ca](https://github.com/r-lib/httr2/pull/846/commits/398b4ca70dd93aad6a093341a8b64501485e64b6).

PR #847, opened 2026-06-17
forward `token_params` on token refresh

simonpcouch:
In service of https://github.com/tidyverse/ellmer/pull/1024.

With these changes, when a token refresh is needed, the OAuth flow will be retriggered when the token expires. Without them, users would see this after an hour:

```
Error in `req_perform_connection()` at ellmer/R/httr2.R:55:5:
! HTTP 401 Unauthorized.
ℹ Unauthorized: Invalid or expired token
```

...with no (exported) helper to reset the token.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # httr2 (development version)
 
-* OAuth token refresh now forwards the `scope` and `token_params` from the original flow, so servers that vary token contents by scope issue an equivalent token on refresh rather than a narrower one.
+* OAuth token refresh now forwards `token_params` from the original flow, so extra token-endpoint parameters (e.g. a `scope` required on the token exchange) are sent on refresh as well as on the initial token request.
 * New `oauth_server_metadata()` discovers an OAuth/OpenID Connect issuer's endpoints from its `.well-known` metadata document (#845).
 * `oauth_flow_auth_code()` now correctly uses the same redirect URI for both authorization and token requests when using the default localhost redirect URL (@pedrobtz, #829).
 * `last_response_json()` now works with content-types that end with `+json`, 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # httr2 (development version)
 
+* OAuth token refresh now forwards the `scope` and `token_params` from the original flow, so servers that vary token contents by scope issue an equivalent token on refresh rather than a narrower one.
 * New `oauth_server_metadata()` discovers an OAuth/OpenID Connect issuer's endpoints from its `.well-known` metadata document (#845).
 * `oauth_flow_auth_code()` now correctly uses the same redirect URI for both authorization and token requests when using the default localhost redirect URL (@pedrobtz, #829).
 * `last_response_json()` now works with content-types that end with `+json`, 
--- a/R/oauth.R
+++ b/R/oauth.R
@@ -50,7 +50,15 @@ auth_oauth_token_get <- function(cache, flow, flow_params = list()) {
       token <- exec(flow, !!!flow_params)
     } else {
       token <- tryCatch(
-        token_refresh(flow_params$client, token$refresh_token),
+        token_refresh(
+          flow_params$client,
+          token$refresh_token,
+          # Forward the scope and token params from the original flow so that
+          # servers which key token contents off scope (rather than the
+          # originally granted scope) issue an equivalent token on refresh.
+          scope = flow_params$scope,
+          token_params = flow_params$token_params %||% list()
+        ),
         httr2_oauth = function(cnd) {
           # If refresh fails, try to auth from scratch
           exec(flow, !!!flow_params)
--- a/R/oauth.R
+++ b/R/oauth.R
@@ -53,7 +53,6 @@ auth_oauth_token_get <- function(cache, flow, flow_params = list()) {
         token_refresh(
           flow_params$client,
           token$refresh_token,
-          scope = flow_params$scope,
           token_params = flow_params$token_params %||% list()
         ),
         httr2_oauth = function(cnd) {
--- a/R/oauth.R
+++ b/R/oauth.R
@@ -53,9 +53,6 @@ auth_oauth_token_get <- function(cache, flow, flow_params = list()) {
         token_refresh(
           flow_params$client,
           token$refresh_token,
-          # Forward the scope and token params from the original flow so that
-          # servers which key token contents off scope (rather than the
-          # originally granted scope) issue an equivalent token on refresh.
           scope = flow_params$scope,
           token_params = flow_params$token_params %||% list()
         ),
--- a/tests/testthat/test-oauth.R
+++ b/tests/testthat/test-oauth.R
@@ -55,19 +55,14 @@ test_that("can refresh to get new token", {
   expect_equal(cache$get(), new_token)
 })
 
-test_that("refresh forwards scope and token_params from the flow", {
+test_that("refresh forwards token_params from the flow", {
   client <- oauth_client("test", "http://example.org/test")
   cache <- cache_mem(client)
   cache$set(oauth_token("123", refresh_token = "456", expires_in = -60))
 
   local_mocked_bindings(
-    token_refresh = function(
-      client,
-      refresh_token,
-      scope = NULL,
-      token_params = list()
-    ) {
-      oauth_token("789", scope = scope, token_params = token_params)
+    token_refresh = function(client, refresh_token, token_params = list()) {
+      oauth_token("789", token_params = token_params)
     }
   )
 
@@ -80,7 +75,6 @@ test_that("refresh forwards scope and token_params from the flow", {
       token_params = list(resource = "example")
     )
   )
-  expect_equal(token$scope, "read")
   expect_equal(token$token_params, list(resource = "example"))
 })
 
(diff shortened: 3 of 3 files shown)


hadley:
This mildly scares me because I have no idea how this will affect existing OAuth providers. Could you make claude do some analysis?

simonpcouch:
> This mildly scares me because I have no idea how this will affect existing OAuth providers. Could you make claude do some analysis?

Agreed, especially re: the first round of changes. Had some back-and-forth and was able to simplify this on the way to implementing silent refresh. The justification from the Claude instance that implemented after some-back-and-forth:

> RFC 6749 §3.2 defines a single token endpoint for both jobs: "The token endpoint is used by the client to obtain an access token by presenting its authorization grant **or refresh token**." A refresh is just another request to that same endpoint — §6 has the client POST `grant_type=refresh_token` there, alongside the same `scope` parameter available on the initial grant. So `token_params`, which httr2 documents as extra parameters for the token request, by definition apply to a refresh too: it *is* a token request. Forwarding them is the consistent behavior; silently dropping them only on refresh is the anomaly this fixes.
> 
> The blast radius is correspondingly narrow. `token_params` defaults to `list()`, and the overwhelming majority of flows never set it — for those, the refresh request is byte-for-byte unchanged, so no existing provider can be affected. The behavior changes *only* when a caller explicitly attached token-endpoint parameters, and in that case the change simply honors the parameters they already asked to send on token requests. (Note this PR no longer forwards the `scope` *argument* — only `token_params` — precisely because §6 says an omitted `scope` "is treated as equal to the scope originally granted by the resource owner," so re-sending it is unnecessary for conformant servers and is the part that could plausibly surprise one. `token_params` carries no such ambiguity.)

A gippety 5.5 adversarial review with a fresh context:

> Yes. Focusing only on the current change, I think it’s technically correct.
> 
> `token_params` means “extra parameters for the token endpoint,” and refresh is also a token-endpoint request. Forwarding those params makes refresh behave more like the original token exchange. OAuth explicitly allows `scope` on refresh requests, with the constraint that it cannot exceed the originally granted scope, and extension token-endpoint parameters are part of OAuth’s model.
> 
> The compatibility risk is real but narrow: a strict provider could reject a refresh request if a user supplied a `token_params` value that was valid for the original grant but invalid for `grant_type=refresh_token`. That would likely surface as `invalid_request`, which OAuth defines for malformed requests or unsupported parameter values.
> 
> I don’t think that risk is very substantive because this only affects callers who opted into non-empty `token_params`; default OAuth use is unchanged. Also, some token params are explicitly meaningful on refresh. For example, RFC 8707’s `resource` parameter applies to access-token requests for all grant types, and its examples include a refresh-token request carrying `resource`.
> 
> The first draft that forwarded top-level `scope` was more concerning: omitting `scope` on refresh means “same as originally granted,” while sending it can ask for a narrower/different token. The current version avoids that broad behavior change unless the caller explicitly put `scope` in `token_params`.

hadley:
Could you please make claude do a github code search to identify existing uses of `token_params`?

(Also worth double checking that posit.ai is doing the right thing, since obviously no other oauth provider has needed this)

simonpcouch:
> Could you please make claude do a github code search to identify existing uses of token_params?

Some evidence that others have had to hack around this:

> `r-lib/httr2#847` ("forward `token_params` on token refresh", by you, open) makes `auth_oauth_token_get()` pass the flow's original `token_params` through to `token_refresh()`. Before, extra token-endpoint params (e.g. a `resource` or `scope` required on the token exchange) were sent on the *initial* token request but dropped on refresh — so refreshable flows would 401 after the token expired. It's in service of `tidyverse/ellmer#1024`.
>
> So the "usages" that matter are downstream packages passing `token_params` into an httr2 OAuth flow — and specifically the **refreshable** flows (`oauth_flow_auth_code` / `req_oauth_auth_code`, device flow), since those are the ones this PR actually changes. Client-credentials flows re-run the whole flow on expiry and never had a refresh step, so they're unaffected.
>
> ## GitHub Code Search results
>
> Directly relevant (refreshable flow + `token_params`) — these benefit from the fix:
>
> - **`lawremi/wizrd`** (`R/oauth.R`) — nearly the exact motivating case:
>   ```r
>   params <- as.list(c(resource = auth@resource))
>   httr2::req_oauth_auth_code(req, auth@client, auth@auth_url, auth@scope,
>                              auth_params = params, token_params = params, ...)
>   ```
>   A `resource` param sent on both auth and token; previously it'd be dropped on refresh.
>
> - **`databrickslabs/brickster`** (`R/request-helpers.R`) — forwards a stored `token_params` into a flow:
>   ```r
>   ..., scope = oauth_client$scope, token_params = oauth_client$token_params
>   ```
>
> Uses `token_params` but **not affected** (client-credentials flow, no refresh step — re-fetches fully each time):
>
> - `R-ArcGIS/arcgisutils` — `oauth_flow_client_credentials(cln, token_params = list(expiration = ...))`
> - `benrwoodard/adobeanalyticsr` and `benrwoodard/cjar` — `oauth_flow_client_credentials(..., token_params = > c(grant_type=..., client_id=...))`
>
> Not related (false positive): `lukakoning/shinyOAuth` uses its own `extra_token_params` slot in a hand-rolled OAuth  implementation, not httr2's `token_params`.
>
> The `wizrd` case is the strongest external confirmation that the bug bites real users: an auth-code flow carrying a
`resource` token param, which is precisely the kind of value that needs to survive a refresh.

I pressed to further ask for suspected breakages in any of those uses and could not find any, but did not run a revdepcheck.

<details><summary>Here's a write-up</summary>
<p>


From the hits I inspected:

- `lawremi/wizrd`: no break expected; likely benefits. It passes:

```r
auth_params = params, token_params = params
```

where `params` is:

```r
list(resource = auth@resource)
```

That is exactly the RFC 8707 case where refresh should carry `resource`.

- `ccs-amsterdam/amcat4r`: no break from this PR. It already calls `httr2:::token_refresh()` directly with:

```r
token_params = list(resource = server, refresh_mode = ..., session_type = "api_key")
```

The PR changes cached automatic refresh, not direct `token_refresh()` behavior.

- `tidyverse/ellmer`: no break; fixes the immediate Posit AI refresh symptom by forwarding:

```r
token_params = list(scope = "prism")
```

- `daveneti/brapi-r6`: possible downstream risk, but no concrete break visible. It exposes arbitrary `token_params` through `req_oauth_auth_code()`. If users pass grant-specific params like `code`, `redirect_uri`, or `grant_type`, refresh could break. Defaults are empty.

- `HIBio/docusignapi`: risky pattern, but probably not affected by this PR because it uses direct `oauth_flow_auth_code()`, not cached request auth:

```r
token_params = list(grant_type = "authorization_code")
```

If this were used through automatic refresh, it could break by sending `grant_type = "authorization_code"` on a refresh request.

- `R-ArcGIS/arcgisutils`: no break expected. It uses:

```r
oauth_flow_client_credentials(..., token_params = list(expiration = expiration))
```

and refresh is a separate direct call without `token_params`.

- `databrickslabs/brickster`: no break expected; visible `token_params` are empty.

- `cran/genesysr`: no break expected; visible `token_params` are empty and direct client-credentials flow.

- `kandrsn99/eveRwrapper`: no break expected; visible `token_params` are empty.

- `benrwoodard/cjar`, `SpeciesFileGroup/Rtaxonworks`, `paws-r/paws`: false positives for this question; they use `httr` or unrelated variables, not httr2 OAuth `token_params`.

- `cran/rtweet`: has odd direct-flow patterns, but not a clear PR break. This is suspicious existing code:

```r
token_params = list(client_secret = client$secret,
                    client_type = "third_party_app",
                    grant_type = "refresh_token")
```

but it’s on direct `oauth_flow_client_credentials()`, not automatic refresh. The cached refresh path uses:

```r
token_params = list(client_id = client$id)
```

which is plausibly intentional.

Net: the only concrete break class I’d name is packages/users putting grant-specific core OAuth fields into `token_params`. I did not find a public package where the PR clearly breaks automatic refresh out of the box.

</p>
</details>


---

> "Also worth double checking that posit.ai is doing the right thing, since obviously no other oauth provider has needed this"

Totally. I'd probably describe this as:

* The client-side change is a spec-blessed hardening (that seems to address an issue others have come across in the wild).
* The analogous server-side change is a fix for a genuine bug.

On the httr2 side, RFC 8707 §2.2 (Resource Indicators), emphasis my own:

 > "When the resource parameter is used on an access token request made to the token endpoint, **for all grant types**, it indicates the target service or protected resource where the client intends to use the requested access token."

On the lucid-auth side, there is indeed a separate issue that would address this on its own:

> `login.posit.cloud`'s token endpoint only mints a gateway-authorized token when `scope=prism` is sent on the token exchange itself, and it doesn't carry that scope forward when a token is later refreshed. RFC 6749 §6 says it should: an omitted `scope` on a refresh request "is treated as equal to the scope originally granted by the resource owner." If the authorization server honored that default, a refresh would reissue a properly-scoped token on its own, and the httr2 change would not be necessary.

Your call on whether this is an ellmer-side hotfix, this httr2 change, a server-side bug fix, or any combination of them. I can file the server-side changes in the coming weeks, but I don't know what their appetite for these changes or deployment / release cycle looks like.

hadley:
Ok, I'm convinced 😄 I think we should fix here and you should encourage the posit.ai team to also fix.

jonthegeek:
FWIW, I wouldn't be surprised if there are a lot more cases where this would be helpful but they aren't showing up because people gave up on getting refresh to work. Just because it shouldn't work that way doesn't mean it doesn't often work that way, in my experience with APIs.

PR #853, opened 2026-06-18
Include request in mocked and cached responses

### LINKED ISSUE #841: Include `resp$request` in mocked responses
Normal responses get the `request` field via `create_response()`, but mocked responses only get the `request` field if they're errors (via `handle_resp()`). This means the second (loaded from mock) response doesn't exactly match the "real" response (in this example, loaded via httptest2, which *might* be the proper place to "fix" this, I'm definitely open to that discussion).

The `is_error(resp)` branch in `handle_resp()` currently sets `resp$request <- req`. If that were either set outside the `if` or also set in the final `else`, mocked responses would have the same `request` objects as their unmocked cousins, regardless of the mock-loader that's being used.

I got the expected result locally by adding `resp$request <- req` in that final `else`, and could PR that easily, but I want to check if you have a preference of how to handle this (if at all).

hadley:
Real responses carry the originating request in `resp$request` via `create_response()`, but mocked and cached responses only did so when they were errors. Set `resp$request <- req` in the success branches of `handle_resp()` and the pooled request handler so all responses match.

Fixes #841


jonthegeek (approved):
Thanks! I don't know if I would have done the proper search to realize that there's another `handle_response()` that also needed the fix, so I'm glad I didn't PR directly!

Would it make sense to abstract out the shared parts of the two `handle_response()`s to ensure that they don't diverge? Without digging into the `PooledRequest` method, I *assume* there's a reason for the differences they currently have, but I also assume you don't want them to diverge in unplanned ways.

hadley:
IIRC it's not possible to reuse any code because the flow is so different.

PR #855, opened 2026-06-18
Reuse `claim` for jwt_sig client auth in bearer JWT flow

### LINKED ISSUE #825: req_oauth_bearer_jwt requires claim twice

hadley:
`req_oauth_bearer_jwt()` uses a JWT as an authorization grant (RFC 7523 §2.1), while `oauth_client(auth = "jwt_sig")` uses a JWT to authenticate the client (§2.2). When both were combined, the same `claim` had to be supplied twice.

Now the bearer JWT flow reuses its `claim` for the client assertion when the client doesn't carry its own, so `oauth_client(auth = "jwt_sig")` no longer requires a `claim` in `auth_params` at creation time; the requirement is instead enforced at request time.

Fixes #825

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,7 @@
 * New `oauth_server_metadata()` discovers an OAuth/OpenID Connect issuer's endpoints from its `.well-known` metadata document (#845).
 * `oauth_client()` gains a `metadata` argument: pass the result of `oauth_server_metadata()` and the client carries all of the server's endpoints, so the OAuth flows pick them up automatically instead of threading them into each call (#846).
 * `oauth_flow_auth_code()` now correctly uses the same redirect URI for both authorization and token requests when using the default localhost redirect URL (@pedrobtz, #829).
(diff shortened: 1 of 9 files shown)


hadley:
I don't fully understand JWT but it seems like this makes sense to me and both Claude and codex think it looks good.

PR #859, opened 2026-07-01
Ignore `NA`s when pruning cache files

hadley:
`file.info()` returns `NA` for files it can't stat, which happens when a cache file is deleted (e.g. by another process, or an OS temp cleaner) between when `cache_info()` lists it with `dir()` and when it stats it. These `NA`s propagated into `to_remove`:

- `if (any(to_remove))` errors with "missing value where TRUE/FALSE needed" when there's an `NA` but no `TRUE`.
- Where there is a `TRUE`, `info[!to_remove, ]` turns `NA`-indexed rows into all-`NA` rows, corrupting later pruning passes.

`cache_prune_files()` now coerces `NA` entries in `to_remove` to `FALSE`, leaving such files in place.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httr2 (development version)
 
+* `req_cache()` pruning no longer errors if a cache file is deleted while the cache is being pruned.
+
 # httr2 1.2.3
 
 * Mocked and cached responses now include the originating request in `resp$request`, just like real responses (#841).
--- a/R/req-cache.R
+++ b/R/req-cache.R
@@ -191,6 +191,8 @@ cache_info <- function(path, pattern = "\\.rds$") {
 }
 
 cache_prune_files <- function(info, to_remove, why, debug = TRUE) {
+  to_remove[is.na(to_remove)] <- FALSE
+
   if (any(to_remove)) {
     if (debug) {
       cli::cli_text("Deleted {sum(to_remove)} file{?s} that {?is/are} {why}")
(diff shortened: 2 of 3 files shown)


jeroen:
What leads to this is `info` being a zero-row data frame:

```r
Browse[1]> info
[1] name  size  mtime
<0 rows> (or 0-length row.names)

Browse[1]> info$mtime + max$age
POSIXct of length 0

Browse[1]> info$mtime + max$age < Sys.time()
[1] NA
```

hadley:
Hmmmm, but

``` r
.POSIXct(numeric())
#> POSIXct of length 0
.POSIXct(numeric()) < Sys.time()
#> logical(0)
```

<sup>Created on 2026-07-01 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

jeroen:
Okay so here is the full reproducer for completeness:

```r
unlink(tools::R_user_dir("gh", which = 'cache'), recursive = TRUE)

library(messydates) # Breaks date comparisons!
gh::gh('/repos/cran/manydata')
```

The bug is in the `messydates` package which when loaded changes behavior of date comparisons:


```r
numeric() < Sys.Date()
# logical(0)

library(messydates)
numeric() < Sys.Date()
# [1] NA
```

hadley:
Seems like this isn't httr2's problem to fix 😄

PR #861, opened 2026-07-10
Allow configurable OAuth expiry margins

### LINKED ISSUE #860: Allow configuring when OAuth tokens are treated as expired
### Problem

`httr2` currently treats an OAuth token as expired five seconds before `expires_at`. Some resource servers require a larger safety margin. In particular, Azure Databricks rejects tokens with 30 seconds or less remaining; the Databricks Python SDK therefore uses a 40-second margin.

This leaves a window where `req_oauth_client_credentials()` and `req_oauth_auth_code()` can reuse a token that `httr2` considers valid but the resource server rejects. The normal OAuth reauthentication path does not help here because Databricks returns 403 rather than a 401 response with `WWW-Authenticate: error="invalid_token"`.

This surfaced in databrickslabs/brickster#254. A downstream workaround can inspect and clear the request cache, but that relies on private request internals and can discard a U2M refresh token, causing the original authorization flow to run instead of refreshing.

References:

- Current five-second margin: https://github.com/r-lib/httr2/blob/main/R/oauth-token.R#L69-L74
- Databricks Python SDK 40-second margin: https://github.com/databricks/databricks-sdk-py/blob/main/databricks/sdk/oauth.py#L110-L119
- Original brickster report: https://github.com/databrickslabs/brickster/pull/254

### Proposed API

Add an argument such as `expiry_margin = 5` (in seconds) to the request-based `req_oauth_*()` helpers and thread it through the existing `token_has_expired(token, delay = ...)` check. Keeping five seconds as the default would preserve current behavior while allowing clients such as brickster to request a 40-second margin without manipulating the cache.

I am happy to submit a PR with tests and documentation if this API direction seems reasonable.

zacdav-db:
## Summary

- Add `expiry_margin` to the request-based `req_oauth_*()` helpers.
- Increase the default OAuth expiry margin from 5 to 30 seconds.
- Pass the margin through the existing expiry and refresh path, preserving refresh-token behavior.

## Motivation

Some resource servers reject tokens shortly before their stated expiry. Databricks, for example, returns HTTP 403 when a token has 30 seconds or less remaining and does not send the RFC 6750 `WWW-Authenticate: error="invalid_token"` challenge that triggers httr2 automatic reauthentication.

A configurable margin lets clients add provider-specific headroom without inspecting or clearing httr2 private cache state. The 30-second default follows the direction discussed in #860; clients can select a larger value when they need additional clock-skew tolerance.

Fixes #860.

## Validation

- `devtools::test(filter = "^oauth")`: 116 passed, 1 credential-gated test skipped.
- `devtools::test()` with `rlang` 1.3.0: 1,300 passed, 2 credential-gated tests skipped.
- CI-generated hash snapshots from run 29094916404 were reviewed and accepted.
- `pkgdown::check_pkgdown()`: no problems found locally.
- `devtools::check()`: 0 errors, 0 warnings, 1 environment-only note (`unable to verify current time`).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,7 @@
 * `req_body_form()` and `req_url_query()` no longer error with "C stack usage is too close to the limit" when given very long string values (#805).
 * `req_cache()` no longer errors when a request is first performed with `path` then later without it (#840).
 * `req_error()` is now applied to responses retrieved from the cache, so a custom `is_error` callback is respected on cache hits (#806).
+* `req_oauth_*()` gains an `expiry_margin` argument to control how early cached OAuth tokens are treated as expired; the default margin increases from 5 to 30 seconds (@zacdav-db, #860).
 * `req_oauth_bearer_jwt()` now uses its `claim` as the basis for a separate client assertion when the `client` also authenticates with `auth = "jwt_sig"`, so you no longer need to supply the claim twice. As a result, `oauth_client(auth = "jwt_sig")` no longer requires a `claim` in `auth_params` at creation time (#825).
 * `req_oauth_device()` gains a `pkce` argument to enable Proof Key for Code Exchange, matching `oauth_flow_device()` (#834).
 * `req_throttle()` can now enforce multiple rate limits at once: supply a vector to `capacity` (and `fill_time_s`) to create one token bucket per limit, and each request must satisfy all of them (#555).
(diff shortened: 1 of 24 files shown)


zacdav-db:
Updated the snapshots for `rlang` 1.3.0 in [`6f7b415`](https://github.com/r-lib/httr2/commit/6f7b4151).

- Downloaded and reviewed the snapshots from the failing CI run.
- Confirmed the four changes were expected hash-only updates.
- Accepted them with `testthat::snapshot_accept()`.

The full local test suite now passes.

zacdav-db:
Should `oauth_token_cached()` also gain an `expiry_margin` argument for consistency? It uses the same cached-token logic and inherits the new 30-second default, but callers currently cannot override it.

hadley:
Yeah, it should. Let me fix the hashes in a separate PR.

hadley:
Thanks!

PR #869, opened 2026-08-23
Include query string in `req_dry_run()` output

### LINKED ISSUE #868: req_dry_run() omits the query string from its output
## Issue

req_dry_run() doesn't include the URL query string in either its printed output or its invisibly-returned list, even though the request clearly contains one. This contradicts the function's stated purpose of showing "exactly what httr2 will send to the server". It appears to be a regression as the [httr2 1.0.0 blog post](https://tidyverse.org/blog/2023/11/httr2-1-0-0/) shows query strings being printed (HEAD /?param=value HTTP/1.1).

## Reprex

```r
library(httr2)

request("https://example.com") |>
  req_url_path("/search") |>
  req_url_query(q = "hello", limit = 10) |>
  req_dry_run()
#> GET /search HTTP/1.1
#> ...
#> (no ?q=hello&limit=10 anywhere in the output)

# By contrast, the request object itself has the full URL:
request("https://example.com") |>
  req_url_path("/search") |>
  req_url_query(q = "hello", limit = 10) |>
  (\(req) req$url)()
#> [1] "https://example.com/search?q=hello&limit=10"
```

## Cause
`curl::curl_echo()` splits the request into separate `path` (`PATH_INFO`) and `query` (`QUERY_STRING`) fields, following Rook/httpuv convention; the query string is captured, in `resp
query`, but `req_dry_run()` only ever references `resp_path`:

```r
resp <- curl::curl_echo(handle, progress = FALSE)
...
cli::cat_line(resp$method, " ", resp$path, " HTTP/1.1")   # resp$query is never used
...
invisible(list(
  method = resp$method,
  path = resp$path,                                        # query not returned either
  body = resp$body,
  headers = headers_flatten(headers, redact = redact_headers)
))
```

## Possible Fix

Re-attach `resp$query` to the printed request line when non-empty, and include it in the returned list, e.g.:

```r
path <- resp$path
if (nzchar(resp$query)) {
  path <- paste0(path, "?", resp$query)
}
cli::cat_line(resp$method, " ", path, " HTTP/1.1")
```
(with a corresponding addition to the invisible return value).

## Session Info

```r
R version 4.6.1 (2026-06-24 ucrt)
Platform: x86_64-w64-mingw32/x64
Running under: Windows 11 x64 (build 26200)

attached base packages:
[1] stats     graphics  grDevices datasets  utils     methods   base

other attached packages:
[1] httr2_1.3.0

loaded via a namespace (and not attached):
 [1] compiler_4.6.1  magrittr_2.0.5  R6_2.6.1        cli_3.6.6
 [5] promises_1.5.0  later_1.4.8     tools_4.6.1     otel_0.2.0
 [9] glue_1.8.1      curl_7.1.0      Rcpp_1.1.2      httpuv_1.6.17
[13] lifecycle_1.0.5 rlang_1.3.0     renv_1.2.3
```

wikisqueaks:
`curl::curl_echo()` splits the request into separate `path` and `query`
fields, but only `path` was used, so the printed request line and the
returned list both omitted the query. The httr2 1.0.0 blog post shows
`HEAD /?param=value HTTP/1.1`, so this appears to be a regression.

`query` is returned verbatim, including the leading `?` that httpuv puts
in `QUERY_STRING` — happy to strip it if you'd rather it matched `path`.

Existing snapshots are unchanged, since they all use query-less URLs.

Fixes #868.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # httr2 (development version)
 
+* `req_dry_run()` now shows the URL's query string in the printed request line and returns it in the `query` element of its result; previously both omitted it (@wikisqueaks, #868).
+
 # httr2 1.3.0
 
 * Fixed OAuth token cache pruning so that it actually matches the encrypted `.rds.enc` files written to disk; previously the pruning pattern only matched an unencrypted `.rds` file that was never created, so cached tokens were never automatically deleted regardless of age.
--- a/R/req-dry-run.R
+++ b/R/req-dry-run.R
@@ -26,7 +26,7 @@
 #'     built.
 #' @param pretty_json If `TRUE`, automatically prettify JSON bodies.
 #' @returns Invisibly, a list containing information about the request,
-#'   including `method`, `path`, and `headers`.
+#'   including `method`, `path`, `query`, and `headers`.
 #' @export
 #' @examples
 #' # httr2 adds default User-Agent, Accept, and Accept-Encoding headers
@@ -70,7 +70,7 @@ req_dry_run <- function(
   )
 
   if (!quiet) {
-    cli::cat_line(resp$method, " ", resp$path, " HTTP/1.1")
+    cli::cat_line(resp$method, " ", resp$path, resp$query, " HTTP/1.1")
 
     if (testing_headers) {
       # curl::curl_echo() overrides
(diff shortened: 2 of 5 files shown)


wikisqueaks:
The failing check is a `pull_request_target` workflow that can't check out fork code under the new `actions/checkout` default. It won't affect PRs from branches in the repo, only forks. R-CMD-check passes.

hadley:
Thank you!

